PR #2, opened 2019-05-31
Rollup build setup

greglobinski:


abhiaiyer91:
We should probably enforce the same rules as the mansion repo

KyleAMathews:
I've been using https://github.com/developit/microbundle for this sort of thing which simplifies setup.

greglobinski:
> We should probably enforce the same rules as the mansion repo for eslint, babel etc

@abhiaiyer91  I believe I did it with this PR https://github.com/gatsby-inc/gatsby-interface/pull/2#issuecomment-498045345

but do you mean  we should go with the exact setup? The ones on Cloud is for Typescript, so I removed the Typescript stuff from the setup, like for example I changed `parser` to `babel-eslint` in `.eslintrc`. I can revert the changes.

greglobinski:
> I've been using [developit/microbundle](https://github.com/developit/microbundle) for this sort of thing which simplifies setup.

@KyleAMathews 	I will try it

PR #6, opened 2019-06-25
Link component

rizchelwood:
Currently, the links are split into external and internal but I think it would be better if we could share between the two components. 

I'll look into this further tomorrow. @greglobinski let me know if you have any ideas


greglobinski:
> Currently, the links are split into external and internal but I think it would be better if we could share between the two components.
:+1: agree

src/components/Link/ExternalLink.Skeleton.js
```diff
@@ -0,0 +1,32 @@
...
+export const externalLinkPropTypes = {
+  href: PropTypes.string,
+}
+
+const externalLinkSkeletonPropTypes = {
+  ...externalLinkPropTypes,
+  StyledComponent: PropTypes.any,
+}
+
+export const externalLinkDefaultPropTypes = {}
```
greglobinski: do we need this if it's empty?

src/components/Link/BaseLink.js
```diff
@@ -0,0 +1,21 @@
+import { fontSizes, fontFamilies, palette } from "../../utils/presets"
+
+export const baseStyles = {
```
greglobinski: I assume that when we combine the `External` and `Internal` into one component this file will not be necessary, but as a note, I think that we should stay consistent and stay with using `styled-component` way and `template literals` wherever we can. But we can discuss it.

src/components/Link/BaseInternalLink.js
```diff
@@ -0,0 +1,29 @@
...
+
+import InternalLinkSkeleton, {
+  internalLinkPropTypes,
+  internalLinkDefaultPropTypes,
+  SkeletonStyledComponent,
+} from "./InternalLink.Skeleton"
+
+import { baseStyles } from "./BaseLink"
+
+export const StyledInternalBaseLink = styled(SkeletonStyledComponent)`
```
greglobinski: instead of manually reassigning the values we can use 

```
export const StyledInternalBaseLink = styled(SkeletonStyledComponent)`
  ${baseStyles}
`
```

src/components/Link/BaseLink.js
```diff
@@ -0,0 +1,21 @@
...
+    alignItems: 'center',
+    display: 'flex',
+    fontSize: `${fontSizes.xs}`,
```
greglobinski: I think that the link should get `font-size` and `font-family` by inheritance, `font-style` and `font-weight` also, not 100% sure in these cases though.

src/components/Link/BaseLink.js
```diff
@@ -0,0 +1,21 @@
...
+    fontFamily: `${fontFamilies.bodyFontFamily}`,
+    fontStyle: 'normal',
+    fontWeight: 'normal',
+    letterSpacing: '0.25px',
```
greglobinski: if we really need this (I'm not convinced) it needs to be set in relative units `em`

src/components/Link/BaseLink.js
```diff
@@ -0,0 +1,21 @@
+import { fontSizes, fontFamilies, palette } from "../../utils/presets"
+
+export const baseStyles = {
+    alignItems: 'center',
+    display: 'flex',
```
greglobinski: should not it be `inline-flex`?

rizchelwood:
@greglobinski made some big changes, this is ready for review!

greglobinski:
Looks good to me :+1:

rizchelwood:
I'll upgrade the version and publish then merge

PR #25, opened 2019-08-02
(WIP) Add base presets

rizchelwood:
Adds base styles to presets 

Resolves part of https://github.com/gatsby-inc/mansion/issues/3977

### DIFF
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -1,0 +1,80 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import PropTypes from "prop-types"
+
+import {
+  styles,
+  fontSizes,
+  breakpoints,
+  palette,
+  colors,
+} from "../../utils/presets"
+
+const Heading = ({ className, variant = `DEFAULT`, children, ...rest }) => {
+  if (variant === `EMPHATIC`) {
+    return (
+      <h2
+        css={{
+          ...styles.heading,
+          fontSize: fontSizes[`3xl`],
+          fontWeight: `800`,
+          letterSpacing: `-0.02em`,
+          lineHeight: `1`,
+          [`@media (min-width: ${breakpoints.desktop}px)`]: {
+            fontSize: fontSizes[`4xl`],
+          },
+        }}
+        {...rest}
+      >
+        {children}
+      </h2>
+    )
+  }
+
+  if (variant === `SITE`) {
+    return (
+      <h1
+        css={{
+          ...styles.heading,
+          strong: {
+            color: colors.lilac,
+          },
+        }}
+        {...rest}
+      >
+        {children}
+      </h1>
+    )
+  }
+
+  return (
+    <h2
+      css={{
+        ...styles.heading,
+        strong: {
+          color: colors.lilac,
+        },
+        small: {
+          color: palette.grey[500],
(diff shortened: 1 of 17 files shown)


src/components/Heading/Heading.js
```diff
@@ -0,0 +1,80 @@
...
+
+import {
+  styles,
+  fontSizes,
+  breakpoints,
+  palette,
+  colors,
+} from "../../utils/presets"
+
+const Heading = ({ className, variant = `DEFAULT`, children, ...rest }) => {
```
greglobinski: Is it intentional that you destructure the `className` and then do not pass it to the inner `tag`

src/components/Heading/Heading.js
```diff
@@ -0,0 +1,80 @@
...
+  if (variant === `EMPHATIC`) {
+    return (
+      <h2
```
greglobinski: I think we should go with prop `as` to render the component as `h1`, `h2`, etc whenever we need.

Besides `as` we can go with `size` and `variant`, this way we will have a flexible component. Last thing for such a basic component all the styles should be in `styles` object. Same way as here https://github.com/gatsby-inc/gatsby-interface/pull/26

we can discuss this on our pair tomorrow

src/components/Heading/Heading.js
```diff
@@ -0,0 +1,80 @@
...
+    >
+      {children}
+    </h2>
+  )
+}
+
+Heading.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  variant: PropTypes.oneOf([`DEFAULT`, `EMPHATIC`, `SITE`]),
```
greglobinski: the `SITE` value is I think too 'local' we should find more general names

src/components/Heading/Heading.js
rizchelwood: Whoops, `className` should be getting passed to the component. I will fix this

src/components/Heading/Heading.js
rizchelwood: Yes we should definitely discuss these topics more tomorrow.

If we're going to use prop `as`, we need to be consistent in all components. We're currently not using it in other components and have used more of this approach. 

re: `styles object -  I originally had it all in `styles` but then changed it to have it in the component. I kept thinking and not sure if we want to have this is in the base styles rather than just in the component. I think it might get complicated when we start adding variations of base components in the styles. It can get messy

src/components/Heading/Heading.js
rizchelwood: Maybe like section heading? I can't think of a good name for this lol

src/components/Heading/Heading.js
greglobinski: > if we want to have this is in the base styles rather than just in the component

:+1 for styles which are directly related to  the  views of the Preview, but  I think we should first create a general `Heading`, `Text` etc components which should work as a atomic blocks, only on them we can build components which relates to the particular Previev views

src/components/Heading/Heading.js
greglobinski: we can repeat the api from `Button`, the only different for `SITE` is color, we also need the heading in `red` for `Delete Site` card so let's use `tone` for this differentiation.  so default tone for `Heading` will be `NEUTRAL` an so on

PR #26, opened 2019-08-05
Move shared styles into `styles` object [Button]

greglobinski:
- install `storybook` add-on `storybook-readme`
- remove `@storybook/addon-info`
- write `README`s for `Button`
- split `Button` component into 
	- `skeletons/BaseButton` - no-style component which manages rendering proper `tag` depending on passed props 
	- and `cloud/Button` - `BaseButton` with Gatsby Cloud styling
- install `gatsby-desing-tokens`
- import `colors` tokens from `gatsby-desing-tokens` 
- update `colors` usage 
	- from `color: ${palette.grey[500]};`
	- to `color: ${colors.grey[50]};`
- create `button` object in `styles` with shared `Button` styles

Peek 2019-08-05 12-46

TODO:
	- create `PrimaryButton`, `SecondaryButton` ... components for temporary compatibility with the Preview
	- write tests


src/components/cloud/Button/Button.js
```diff
@@ -0,0 +1,47 @@
...
+} from "../../../utils/options"
+
+const Button = ({
+  children,
+  css,
+  DefaultIcon = MdArrowForward,
+  loading,
+  LoadingIcon = MdRefresh,
+  size = `L`,
+  tone = `STANDARD`,
```
rizchelwood: Should tone be `DEFAULT` instead of standard? That's the default value we have for other components

src/components/cloud/Button/Button.js
```diff
@@ -0,0 +1,47 @@
+/** @jsx jsx */
+import { jsx, keyframes } from "@emotion/core"
```
rizchelwood: We can discuss this more tomorrow but I'm wondering if creating a `cloud` folder is a good idea. In the future, we won't want to create a folder for each app that will be using this. Maybe there's a way we can generalize this instead of it being so specific to cloud?

src/components/skeletons/BaseButton/BaseButton.js
```diff
@@ -0,0 +1,103 @@
+import React, { Fragment } from "react"
+import { jsx, keyframes } from "@emotion/core"
```
rizchelwood: Should we include base components in its component folder instead of skeleton? We probably don't want to separate them

src/components/skeletons/BaseButton/BaseButton.js
```diff
@@ -0,0 +1,103 @@
...
+    disabled = false,
+    href,
+    label,
+    DefaultIcon,
+    loading = false,
+    loadingLabel = `Loading`,
+    LoadingIcon,
+    rel,
+    role = `button`,
+    target = `_blank`,
```
rizchelwood: We should add `rel="noopener"` if the target equals `_blank`

src/components/skeletons/BaseButton/BaseButton.js
```diff
@@ -0,0 +1,103 @@
+import React, { Fragment } from "react"
+import { jsx, keyframes } from "@emotion/core"
+import PropTypes from "prop-types"
+import { Link } from "gatsby"
+import { secureTargetBlankLink } from "../../../utils/helpers"
+
+function RewriteChildren({ children, label, DefaultIcon }) {
```
rizchelwood: This sounds a little weird. Maybe there's a better name for it?

src/utils/presets/styles.js
```diff
@@ -0,0 +1,182 @@
...
+        justifyContent: `center`,
+        transition: `background 0.5s, border 0.5s, color 0.5s`,
+        lineHeight: `1`,
+        textDecoration: `none`,
+
+        "&[disabled], &[disabled]:hover": {
+          cursor: `not-allowed`,
+          opacity: !loading ? 0.5 : 0.9,
+        },
+        svg: {
```
rizchelwood: Do we want to include child element styles in base styles? Maybe it should be in the component instead

src/components/cloud/Button/Button.js
greglobinski: You're right :+1, we should look for better name. I was thinking that would be be good to collect all 'styled' components in it's own folder so `skeletons/` and `this_other_folder/` are collateral. Maybe `core/`. My idea is that `skeletons` will contain no-style components, in the `core(?)` we will have components built with `skeletons` + `theme/styles`,

src/components/cloud/Button/Button.js
greglobinski: that's a good question, I choose `STANDARD` to avoid description that "default value of tone prop is default", but I'm not tied to that

src/components/skeletons/BaseButton/BaseButton.js
greglobinski: I think this way, `skeletons` are 100% pure 'logic' component, you get a skeleton apply 
 styles/theme to it and you get a 'styled' component. I called the skeleton button `BaseButton` to differentiate it from styled `Button` component, instead calling it `ButtonSkeleton` because there will be skeletons which could be a base component for more that one 'styled' components. But maybe it's wrong and we should go with a fixed `Skeleton` postfix like did before.

greglobinski: so in the future we can get a skeleton Button/BaseButton apply to them styles/theme completely different than the one we use for Cloud and get a separate set of 'styled' components build on the same set of 'skeletongs'

src/components/skeletons/BaseButton/BaseButton.js
greglobinski: :+1: Absolutely, this component exists only for DRY. `UpdatedChildren`? or maybe instead of using `<Component />` syntax I should switch to `{updateChildren()}` syntax?

src/utils/presets/styles.js
greglobinski: I think we can treat `svg` icon as a internal feature of `Button`, we style it according to the style/tone of the parent so we do not need to exclude it from the here.

rizchelwood:
@greglobinski do you want to merge your pr first? Then I can get access to tones, sizes, etc

PR #34, opened 2019-08-21
add filestack upload component

gillkyle:
This adds a component for use with `filestack-react` for uploading files.

Single, or multi file uploads are configurable through props, as well as overriding the components that are used and the options that `filestack-js` uses.

Here's a demo (yes I know it looks like trash, I was going for functional as per Abhi's request 😎 ):

https://gatsbyjs.slack.com/files/UGZJE0ABY/FMDKPCJGZ/filestack.mp4

### DIFF
--- a/__stories__/FileUpload.stories.js
+++ b/__stories__/FileUpload.stories.js
@@ -1,0 +1,60 @@
+import React from "react"
+
+import { storiesOf } from "@storybook/react"
+
+import FileUpload from "./FileUpload"
+
+storiesOf(`Single File Upload`, module)
+  .add(`Default`, () => {
+    const setFieldValue = props => {
+      console.log(props)
+    }
+    return (
+      <>
+        <FileUpload name="file" setFieldValue={setFieldValue} />
+      </>
+    )
+  })
+  .add(`Multi File Upload`, () => {
+    const setFieldValue = props => {
+      console.log(props)
+    }
+    return (
+      <>
+        <FileUpload
+          multi={true}
+          name="file"
+          setFieldValue={setFieldValue}
+          fileTypes={[`image/jpeg`]}
+        />
+      </>
+    )
+  })
+  .add(`File Upload with Custom Components`, () => {
+    const setFieldValue = props => {
+      console.log(props)
+    }
+    return (
+      <>
+        <FileUpload
+          multi={false}
+          name="file"
+          setFieldValue={setFieldValue}
+          fileTypes={[`image/jpeg`]}
+          CustomButtonComponent={({ onPick }) => (
+            <button type="button" onClick={onPick}>
+              Custom Button
+            </button>
+          )}
+          CustomPreviewComponent={({ file, index, removeFile }) => (
+            <div>
+              <span>{file.filename}</span>
+              <button type="button" onClick={() => removeFile(index)}>
+                X
+              </button>
+            </div>
+          )}
+        />
+      </>
+    )
+  })
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "filestack-react": "^3.1.0",
     "gatsby-design-tokens": "^1.0.0"
   },
   "peerDependencies": {
--- a/src/components/FileUpload/FileUpload.js
+++ b/src/components/FileUpload/FileUpload.js
@@ -1,0 +1,186 @@
+import React, { useState, useEffect } from "react"
+import PropTypes from "prop-types"
+import ReactFilestack from "filestack-react"
+
+import { Button } from "../core/Button"
+import UploadPreview from "./UploadPreview"
+
+const defaultFilestackOptions = {
+  accept: [`image/jpeg`, `image/png`],
+  maxSize: 10 * 1024 * 1024, // 10MB
+  fromSources: [`local_file_system`, `url`, `googledrive`, `dropbox`, `github`],
+}
+
+const FileUpload = ({
+  actionOptions,
+  CustomButtonComponent,
+  CustomPreviewComponent,
+  fileTypes,
+  multi = false,
+  name,
+  setFieldValue,
+  ...rest
+}) => {
+  const [files, setFiles] = useState([])
+
+  const removeFile = index => {
+    if (index === undefined) {
+      throw new Error(
+        `you must provide an index (from props) to the removeFile method in a CustomPreviewComponent, ie removeFile(index)`
+      )
+    }
+    if (index >= files.length || index < 0) {
+      throw new Error(
+        `remove file failed because the index for the removed file is outside of range`
+      )
+    }
+    files.splice(index, 1)
+    setFiles([...files])
+  }
+
+  const addFiles = uploadedFiles => {
+    if (multi) {
+      // set multiple files to state
+      const allFiles = [...files, ...uploadedFiles]
+      setFiles([...allFiles])
+      const fileUrls = allFiles.map(file => file.url)
+      setFieldValue(name, [...fileUrls])
+    } else {
+      // overwrite a single file in state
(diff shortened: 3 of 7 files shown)


__stories__/FileUpload.stories.js
```diff
@@ -0,0 +1,60 @@
+import React from "react"
```
rizchelwood: Could you place this story in the FileUpload component? 

We originally had stories in a separate folder but are now putting them with their components

__stories__/FileUpload.stories.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { storiesOf } from "@storybook/react"
+
+import FileUpload from "./FileUpload"
+
+storiesOf(`Single File Upload`, module)
+  .add(`Default`, () => {
+    const setFieldValue = props => {
+      console.log(props)
```
rizchelwood: Should these be logged in the storybook console - `@storybook/addon-console` - https://www.npmjs.com/package/@storybook/addon-console

package.json
```diff
@@ -20,6 +20,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "filestack-react": "^3.1.0",
```
rizchelwood: I'm not sure but do you know if this needs to be a peer dependency?

src/components/FileUpload/UploadPreview.js
```diff
@@ -0,0 +1,33 @@
+import React from "react"
+
+import { Button } from "../core/Button"
+
+const UploadPreview = ({ file, removeFile, index }) => (
+  <div style={{ display: `flex`, alignItems: `center` }}>
```
rizchelwood: Can you change these `style` props to `css`? You'll also have to add the `jsx` pragma to this file and import it from `@emotion/core`

rizchelwood: Also do you know if this component will be a base component/used in future components? If so, we'd probably want to move it to the core folder. If not, this is good 

cc: @greglobinski what do you think?

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+    },
+    status: `Stored`,
+  }
+  useEffect(() => {
+    addFiles([{ ...mock }])
+  }, [])
+
+  const isEmpty = !!files.length
+
+  const buttonText = (() => {
```
Elanhant: Can this be rewritten as a simple ternary expression?
```javascript
const buttonText = isEmpty
  ? `Choose ${multi ? `another` : `a different`} file`
  : `Pick file${multi ? `s` : ``}``;
```

src/components/FileUpload/UploadPreview.js
```diff
@@ -0,0 +1,33 @@
...
+    <img
+      style={{
+        height: 100,
+        width: 100,
+        margin: 0,
+        objectFit: `contain`,
+      }}
+      src={file.url}
```
rizchelwood: Sorry this is just picky of me but could you destructure file component at the top of this component `const { url, filename } = file`. There also should be some checks if these keys do not exist that it returns `null` or something so it doesn't break it

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+          key={file.uploadId}
+          file={file}
+          index={index}
+          removeFile={removeFile}
+        />
+      )
+    } else {
+      return (
+        <UploadPreview
+          key={file.uploadId}
```
Elanhant: Seems like all the 4 props passed to `UploadPreview` and `CustomPreviewComponent` are the same. Can we do smth like
```javascript
const previewProps = {
  key: file.uploadId,
  file,
  index,
  removeFile,
};

return CustomPreviewComponent
  ? < CustomPreviewComponent {...commonProps} />
  : <UploadPreview {...commonProps} />;
```

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  const addFiles = uploadedFiles => {
+    if (multi) {
+      // set multiple files to state
+      const allFiles = [...files, ...uploadedFiles]
+      setFiles([...allFiles])
+      const fileUrls = allFiles.map(file => file.url)
+      setFieldValue(name, [...fileUrls])
+    } else {
```
rizchelwood: You can remove this `else` statement. Just set what's inside below the `if` statement

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+  useEffect(() => {
+    addFiles([{ ...mock }])
+  }, [])
+
+  const isEmpty = !!files.length
+
+  const buttonText = (() => {
+    if (isEmpty) {
+      return `Choose ${multi ? `another` : `a different`} file`
+    } else {
```
rizchelwood: You can also get rid of this `else` statement like above

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+
+  const renderButton = onPick => {
+    // use custom component if provided
+    if (CustomButtonComponent) {
+      return (
+        <CustomButtonComponent onPick={onPick}>
+          {buttonText}
+        </CustomButtonComponent>
+      )
+    } else {
```
rizchelwood: re: `else` statement, here too

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+          ...actionOptions,
+          accept: fileTypes && [...fileTypes],
+        }}
+        onSuccess={result => {
+          addFiles(result.filesUploaded)
+        }}
+        customRender={({ onPick }) => renderButton(onPick)}
+        {...rest}
+      />
+      {isEmpty && files.map((file, index) => renderPreview(file, index))}
```
Elanhant: We can just use `isEmpty && files.map(renderPreview)`. It will save us unnecessary function recreations on every render at no additional cost.

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,186 @@
...
+        actionOptions={{
+          maxFiles: multi ? 5 : 1,
+          ...defaultFilestackOptions,
+          ...actionOptions,
+          accept: fileTypes && [...fileTypes],
+        }}
+        onSuccess={result => {
+          addFiles(result.filesUploaded)
+        }}
+        customRender={({ onPick }) => renderButton(onPick)}
```
Elanhant: Why not just make `renderButton` accept all of `customRender` arguments and use `customRender={renderButton}`? It will also help if later we decide to expand the `renderButton` to do more stuff.

package.json
gillkyle: This comment led me down a good learning rabbit hole on when to use peerDependencies and when to just use regular dependencies and _I think_ this could still remain a regular dependency. 

Seems like the idea of a peerDependency is avoid problems when someone is installing `gatsby-interface` in their project so that there isn't a version collision for the packages that are used in (A) the library - `gatsby-interface` and (B) the project where the library is installed - like `cloud`. So if I understand correctly we would want it as a peerDependency if we were going to install `filestack-react` in `cloud` too and keep the versions from having problems. But, we just want it as this as an implementation detail that takes care of the file upload by bringing along a working version of `filestack-react` when you run install.

Happy if someone that gets this better can help me understand it more, but that's what a couple minutes scanning npm and yarn docs made me think.

src/components/FileUpload/UploadPreview.js
gillkyle: I can't think of any other place this would be used 🤔 seems like it's pretty related to the file upload and can't think of a reason why it'd appear by itself.

PR #36, opened 2019-08-25
feat: toast component and hooks

Elanhant:
This is a port of `Toast` component from cloud; the goal was to keep the styles and see if it can be improved. Main changes when compared to the cloud version:
1. The toast itself now can support different types. As an example, I've added an `error` type — it will not disappear on timeout and has `"assertive"` type that makes screen readers read the message immediately (which is the desired behaviour for an "error" alert).
2. `Toast` does not use `dangerouslySetInnerHTML` to display its message anymore. Judging by its use in cloud, seems like its purpose was to render text containing `<strong>`  tags. The same can be achieved with this new component like this: 
```javascript
showSuccessToast(
  <React.Fragment>
    <strong>User settings</strong> updated sucessfully.
  </React.Fragment<
);
``` 
3. `ToastProvider` now accepts a new prop — `closeButtonLabel`, both to provide an accessible label for the toast close button and to support i18n if needed.
4. Added hooks that allow to show toasts of different types:
a. `useShowSuccessToast`
b. `useShowErrorToast`
b. `useShowToast` (used by the previous hooks under the hood)

### DIFF
--- a/src/components/Toast/Toast.js
+++ b/src/components/Toast/Toast.js
@@ -1,0 +1,132 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import React from "react"
+import PropTypes from "prop-types"
+
+import Alert from "@reach/alert"
+import { keyframes, css } from "@emotion/core"
+import { MdDone, MdClose, MdWarning } from "react-icons/md"
+
+import {
+  fontSizes,
+  dimensions,
+  palette,
+  radius,
+  spaces,
+} from "../../utils/presets"
+import { ToastType } from "./constants"
+
+const toastEntryAnimation = keyframes`
+  100% {
+     transform: perspective(1000px) rotateX(0);
+  }
+`
+
+const toastCss = css`
+  align-items: center;
+  animation: ${toastEntryAnimation} 0.5s 0.25s ease forwards;
+  background: ${palette.grey[900]};
+  border-left: 8px solid ${palette.green[500]};
+  border-radius: ${radius.default} ${radius.default} 0 0;
+  color: ${palette.green[50]};
+  display: flex;
+  font-size: ${fontSizes.xs};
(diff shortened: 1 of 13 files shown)


src/components/Toast/Toast.js
```diff
@@ -0,0 +1,132 @@
...
+  [ToastType.SUCCESS]: MdDone,
+  [ToastType.ERROR]: MdWarning,
+}
+
+export default function Toast({
+  id,
+  message,
+  type,
+  onRemove,
+  closeButtonLabel,
```
greglobinski: Please add a default value to this
```
closeButtonLabel = 'Close'
```

src/components/Toast/Toast.js
```diff
@@ -0,0 +1,132 @@
...
+
+const ToastIconByType = {
+  [ToastType.SUCCESS]: MdDone,
+  [ToastType.ERROR]: MdWarning,
+}
+
+export default function Toast({
+  id,
+  message,
+  type,
```
greglobinski: Could you please change the name of the `prop` from `type` to `tone` with possible values [`SUCCESS`, `DANGER`]? Just for consistency with other components  (Button, Heading)

src/components/Toast/Toast.js
```diff
@@ -0,0 +1,132 @@
...
+  onRemove,
+  closeButtonLabel,
+}) {
+  const IconComponent = ToastIconByType[type]
+
+  return (
+    <Alert
```
greglobinski: :+1: Thank you for that

greglobinski:
> should we encapsulate this error type with the Link component into another hook 

@abhiaiyer91 Sure we can

// @Elanhant

greglobinski:
> I say we go forward and fix this later

@abhiaiyer91 Sure :+1:

greglobinski:
Thank you @Elanhant

PR #41, opened 2019-08-29
fix emphasized radio jumping when selected

Elanhant:
Fixes "emphasized" radio pushing its siblings when it is selected.


Elanhant:
The story is just a draft, I'll probably open a separate PR to extend on it later.

PR #43, opened 2019-08-30
Prevent radio labels from causing other options to jump

Elanhant:
Seems like labels for `emphasized` radios had a larger padding applied. I guess it was meant as an additional indication for a selected option, but it also causes the "jumpiness" in the radio group.


rizchelwood:
@greglobinski let's publish this when we publish the changes you're doing

PR #44, opened 2019-09-04
TypeScript POC + CopyButton component

Elanhant:
This PR setups TypeScript so that it is possible to use it to write new components; it also implements `CopyButton` component in TS.

Couple of notes:
1. `CopyButton` is using ES6 default params instead of React's `defaultProps`. This is due to React team intention to deprecate `defaultProps` in the near future: https://github.com/reactjs/rfcs/pull/107
1. I had to override `quotes` rule in ESLint for TS files since the default rule placed backticks in `JSX.IntrinsicElements["button"]` which results in syntax error.
1. Not sure if we should keep `propTypes` for TS components.

TODO list:

- [x] Since all components but `CopyButton` are in JS, I had to turn on the `allowJs` flag in TS config. This does not allow us to generate declaration files during build. A temporary workaround would be to have a "shadow" index.ts file that exports ONLY TypeScript components and use it to generate the declaration file.
- [x] Setup tests with TypeScript
- [x] Test coverage for `CopyButton`
- [x] Readme for `CopyButton`


abhiaiyer91:
i like where this is going!

Elanhant:
I've updated this PR with Jest setup for TypeScript and a couple of tests for `CopyButton`. Let me know if you have any questions!

.storybook/webpack.config.js
```diff
@@ -2,38 +2,48 @@ module.exports = async ({ config, mode }) => {
   // console.dir(config.plugins, { depth: null }) || config;
```
abhiaiyer91: i think we want to preserve single quotes here

rollup.config.js
```diff
@@ -28,8 +32,20 @@ export default {
     babel({
       exclude: `node_modules/**`,
       plugins: [`@babel/external-helpers`],
+      extensions,
     }),
-    resolve(),
+    resolve({ extensions }),
     commonjs(),
+    // TODO The following two plugins should be removed once all src is using TypeScript
```
abhiaiyer91: lets make an issue for this

src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  html[dir="rtl"] & {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+`
+
+export type GetLabelFn = (copied: boolean) => string
+
+// TODO extend Button props instead of JSX.IntrinsicElements
```
abhiaiyer91: make an issue for this

jest.config.js
```diff
@@ -1,6 +1,11 @@
+const { defaults } = require(`jest-config`)
```
DSchau: I honestly never knew about this. Neat!

tsconfig.json
```diff
@@ -0,0 +1,65 @@
+{
+  "compilerOptions": {
+    "skipLibCheck": true,
+    /* Basic Options */
```
DSchau: Nit, but do we need all of these comments? 🤷‍♂

rollup.config.js
Elanhant: An umbrella issue for updating build configuration: #47

src/components/CopyButton/CopyButton.tsx
Elanhant: Here it is #48

tsconfig.json
Elanhant: Oh they were generated by TypeScript during initialization! I think it might be helpful to keep them for now in case people are not familiar with some options, but I'm also totally fine with omitting them.

DSchau: Works for me!

Elanhant:
So I've merged #61 which rewrites `Button` to TypeScript and I've also added a readme for `CopyButton`. Let me know if there's anything you'd like addressed before I merge this and we finally get to write some beautiful TS code 🎉 

I'm also going to work on a separate PR to provide a guide to writing components in TypeScript (hoping to cover all the topics from #64)

PR #49, opened 2019-09-11
Add an alternative variant to integration row content details

rizchelwood:
This allows the user to specify a variant for the integration row content details component

Currently, it only supports details as a row but this allows it to be a column as well. This is for the new designs for cms integrations


src/components/core/IntegrationRow/IntegrationRow.js
```diff
@@ -140,23 +143,37 @@ IntegrationRow.EditButton = ({ children, label = `Connect`, ...rest }) => {
...
+          },
         }}
       >
-        <Heading as="span" variant="LIGHT">
+        <Heading
+          as="span"
+          variant="LIGHT"
+          css={{
+            fontFamily: primaryStyling
+              ? fontFamilies.headerFontFamily
```
greglobinski: Please use for `fontFamily` tokens from `theme/fonts`

```
fontFamily: fonts.header.join(',')
```

PR #51, opened 2019-09-12
Forward Radio and RadioSkeleton refs to radio input

Elanhant:
Now it will be possible to pass `ref` to `Radio` or `RadioSkeleton` allowing for more control over the input.

Most of the changes here are by prettier :)


Elanhant:
> Do we need that already or it's for future case? I'm just curious.

We might already need it as I'm trying to implement a general `RadioCard` component which would need to have an ability to focus or select the input outside of the label area. 

Plus, in my experience it makes sense to have forwarded refs for components that mirror basic HTML controls like inputs, selectors, buttons etc. That way it becomes much easier to later use them in more complex components.

PR #53, opened 2019-09-15
Icons generator

Elanhant:
Introduces some new tools that allow to quickly add new icon components:
1. `IconSkeleton` component which takes care of scaling the icon and applying color to it. It's basically just an SVG element with some default props.
1. New script, `scaffold:icon`, that generates a stub file for icon component and exports this component in `src/components/icons/icons.js`
1. New storybook stories — one that lists all icons and a separate story for each icon component.

There are some default rules based on assumptions made by `IconSkeleton`:
* An icon should be monochromatic. `IconSkeleton` uses [currentColor](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentColor_keyword) as default color for fill (or stroke) in its SVG code. It also should not mix `stroke` and `fill` to draw its paths.
* Any `stroke` or `fill` attributes on elements inside SVG should be removed so that `IconSkeleton` can take care of colors.
* An icon should have a view box of `"0 0 24 24"`

It is still possible to build icons that don't follow the rules above — you can just pass custom values for SVG attributes to `IconSkeleton` and you still have full control over the inner SVG. Ideally we should avoid such cases as it removes consistency which is usually expected from icon sets. 

This PR also contains four new icons build using `scaffold:icon`.

TODO:
- [x] Export icons in `src/index.js`
- [x] Snapshot testing for icons
- [x] Readme
- [x] ~Should we expose `IconSkeleton` via index file? It can be useful if some new icon is needed for an app that uses `gatsby-interface` but we don't have time to add it here.~ (I don't think it's feasible for now)


abhiaiyer91:
Looking good so far

Elanhant:
Updated the PR to complete TODO list. Should be ready for merge in my opinion, let me know if you guys want something addressed.

I've also added an issue to investigate adding more automation to optimize/clean SVG code: #55

PR #54, opened 2019-09-17
Modify bundling and distribution

### LINKED ISSUE #50: [RFC]: Alternate approach for exports and bundling
The purpose of this RFC is to suggest an alternate approach for exports and bundling strategies.
Right now everything is being re-exported from `src/index.js` which means we are generating single bundle for the whole `gatsby-interface`. There are certain things that needs to be considered from a developer's perspective:

1. Everytime a new component is created the person who created the component also needs to remember to re-export it from index.js (one extra hop).
2. Since things are being re-exported the `package.json` should have `sideEffects: false`. This is basically a hint to webpack that the re-exports are considered side effects free and it can go ahead and tree shake because when we re-export everything is an object and webpack without `sideEffects: false` flag will not be able to tree shake since it can't predict what would be required on runtime and so it imports the whole library. Again this is dependant on webpack different tools will have different ways of controlling this.
3. No default exports, only named exports.
4. All the imports have to be in a single line since everything is a named export now
```js
import { Button, TextButton, PrimaryDeleteButton, Card, CardSkeleton} from 'gatsby-interface'
```

There's an alternate approach that we can consider and which is what I'm trying to suggest here. The idea is to basically achieve something like following:
```js
import Button, { TextButton, PrimaryDeleteButton } from 'gatsby-interface/Button'
import Card, { CardSkeleton } from 'gatsby-interface/Card'
```
With this approach there are certain benefits out of the box:

1. Developer's don't have to remember to re-export the components.
2. Tree shaking agnostic of the tools being used by the consumer application (even if it's internal to gatsby).
3. We can take the advantage of named and default exports/imports.
4. All the imports are explicit so there's no worry of what will be imported in the final bundle of the consumer application.

## What all changes will be required to achieve this?
1. Remove `main` and `module` fields from `package.json`.
2. Remove re-exports file `index.js` from `src` directory.
3. Modify `rollup` config.
4. A simple node script to generate `files` array. This can run as pre-publish step or may be as a commit hook.

I'm open to any feedbacks and suggestions about what you think about this 🙂

Also I can take this up and raise a PR to achieve this end to end.

kamleshchandnani:
This PR sets up eslint, pre-commit hooks and removes unused imports

Here's the todo list
- [x] eslint fixes - f93e87b228de0bf962ca5f381bd6a15d93a4d703
- [x] remove `storybook-static` - bfde60f2229ae4c3a0c38438e898fdb5c75233fd
- [x] add `storybook-static` to `.gitignore` - 06440140e9b72b6253d3446b36245564669c1d9e


DSchau:
Hi @kamleshchandnani --

I like your idea here and even though you're splitting it up into commits, it's still a quite noisy PR to review since most are formatting changes introduced in this commit.

If you are going to work through the rollup/bundling stuff I'd urge you to do it in a separate PR.

kamleshchandnani:
@DSchau Sure makes sense. Actually I was planning to do the same but I need changes in this PR to be there since this will become much simpler. But sure I'll raise a separate PR for rollup/bundling things. 

I request if you can review the PR and merge it if things are alright I can quickly jump to the actual things.

kamleshchandnani:
Hey @DSchau 
Just wanted to check if you have any inputs on this PR?

DSchau:
Hi Kamlesh!

Thanks for bumping -- I don't really think this is hugely valuable to get in. It tends to be a bit more churn-y/preference-y rather than a meaningful improvement.

Could you explain why this PR is necessary, and specifically why it's necessary to unblock e.g. rollup and an improved bundling mechanism? It seems like the latter can be done entirely in the absence of the former.

We very much appreciate the work here, as I hope I've made clear, but I don't think this is worth merging at this time!

kamleshchandnani:
Okay maybe some confusion here. The idea of this PR was just to format and have tools like linters in place. The linters were already there but it wasn’t doing their job. Before the codebase grow and things becomes chaotic I thought to take this as the preliminary step and have things in place.

I believe basic tooling not working is equivalent of not having them. 

It won’t take much time to review because it’s just removal of things which is for the better developer experience here onwards. But if you think it’s of no value than I guess I will close this.

DSchau:
@kamleshchandnani -- I think I wasn't as clear as I could've been, so I'm sorry for that miscommunication! First and foremost, I don't think this PR is of "no value."

I presumed you were going to focus on bundling, so under that umbrella, this PR doesn't make a ton of sense.

Going forward, we'd certainly consider reviewing more PRs if you're willing to make them. That being said, it makes it a _ton_ easier to review if they stay focused and generally more scoped to a singular task. Additionally, if you'd be willing to more clearly document what's "not working" when you open a PR (more broadly speaking, what task you're accomplishing and why), that again makes it clearer.

For example, in this PR, it seems like the main change was introducing additional ESLint rules that caught unused imports, and then you removed the imports. However, this wasn't clear or explained, so it's difficult to know that was your intent!

It may seem that when leaning towards closing a PR that something isn't valued, but that's not the case here at all. I hope we can take some of the findings/learnings here and apply them towards future PRs.

Thanks for your time, and thanks for the contribution. Sorry we didn't get this one merged!

Issue #40, opened 2019-08-28
v0.0.35 errors in .com


rizchelwood:
I published `0.0.35` but I'm getting this error in .com

I will keep it at `0.0.34` for now. I haven't been able to dig into this but it looks like it's having a problem with `filestack` with `gatsby` just from a quick glance

It works fine in storybook

Maybe one of you know more on this? cc: @greglobinski @gillkyle 

<img width="884" alt="Screen Shot 2019-08-28 at 3 03 28 PM" src="https://user-images.githubusercontent.com/18426780/63892638-59822900-c9a5-11e9-929b-8bd2e8ed3549.png">

gillkyle:
Hmmmm. I wonder if Filestack is using some browser globals that need to be wrapped in useEffects or something when initialized.

Though I don't think just installing it should cause the whole build to break unless it's used somewhere and so far it isn't. 🤔

rizchelwood:
@gillkyle have you been able to look more into this?

gillkyle:
No, sorry for the wait. I'll try and prioritize it 👍 got more busy starting school.

rizchelwood:
No worries, thanks!

PR #59, opened 2019-09-19
updating filestack to work when gatsby builds

gillkyle:
Based on the errors from #40, updated the filestack component to use [`@loadable/component`](https://github.com/smooth-code/loadable-components) to ignore the filestack import during `gatsby build`.

The only new changes made with this since it was first added are on [this line](https://github.com/gatsby-inc/gatsby-interface/compare/filestack-with-ssr?expand=1#diff-0b5d89c1ee435a092f509cd5e29bc7b3R8), after installing loadable.


package.json
```diff
@@ -23,6 +23,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@loadable/component": "^5.10.2",
```
gillkyle: Should this not be a regular dependency?

rizchelwood: It is a regular dependency?

gillkyle:
Just changed the base branch to `dev` and fixed conflicts, thanks @rizchelwood!

gillkyle:
According to @wardpeet, this is fine to be included as a regular dependency 👍

rizchelwood:
Thanks @gillkyle! This looks good, will merge 👍

PR #60, opened 2019-09-20
Add context to notification to use close functionality by itself

rizchelwood:


greglobinski:
@rizchelwood  Maybe we can use `sessionStorage` to store the user's `close` action/decision and do not show the notification every time the page/screen component is mounted/re-rendered?

rizchelwood:
@greglobinski I think we should let the user use whatever method they need to outside the component. That's what I do for this component in .com. That way it doesn't handle any data in the library and make it specific to only `sessionStorage` or any method for that fact.

greglobinski:
@rizchelwood But you menage the state of the component inside it 
```
const [showNotificationState, setNotificationState] = useState(
...
```
so how you will know that the user (a user of the app closed the notification) outside of `Notification` component? Seems that I do not understand something.

rizchelwood:
You can pass the initial state of the notification so if you don't want to show it, you would pass that prop as false. If the user closes it in the notification, the state will change it to false and not show it

greglobinski:
@rizchelwood Yeah, I get that, but I'm thinking about the case when the user closes the notification (by clicking the close button) and then for whatever reason the page will be reloaded, or user will back to this page after visiting other pages, in such cases the notification will be presented again, right? And I think we should hide the notification for some time, at least it should not be presented only because the user reloads the page.

rizchelwood:
In that case, you could use `localStorage` (or whatever method you use) in the component that's pulling in notification. If there's something in `localStorage` then pass in false to not show it and if there isn't something in `localStorage`, pass in true to show the notification. In the example you're asking about, when they visit the page set something in `storage` then when they visit or reload the page, it won't show. We should do all the data logic in the parent component then just pass down props to either show or hide the component in the library.

greglobinski:
`when they visit the page set something in storage` yeah but visiting is not enough, you have to store something just after the user clicks the Close button, right, and I expect to make that possible, you have to pass to the component a call back function triggered by Close click, right, but I only see that the code update the inner state of the component

rizchelwood:
I think for now, this is good. In the example that Flo had, the notification only shows the first time the user visits a page so this works for that use case. We can update the component as other use cases come in

greglobinski:
> We can update the component as other use cases come in 

sure :+1:

PR #61, opened 2019-10-03
Button component rewritten to TypeScript

Elanhant:
Opening this PR against `copy-button` to reduce the amount of code to review.

In this PR `Button` component is rewritten using TypeScript. I'll try to describe the process thoroughly, but please feel free to ask questions or provide suggestions if anything does not make sense! I'm hoping to compile a TypeScript migration guide for `gatsby-interface`, so feedback is very much appreciated. Also, there are some questions from me at the end of this post :)

### Split responsibilities
Previously, `BaseButton` could be used to display an HTML anchor link or Gatsby `Link`. While this might seem useful at first, I think it brings unneeded complexity, uncertainty and ambiguity to the component's usage. When an engineer imports a component called `Button`, they are more likely to expect it to behave like a button first, and only then to look like one. Therefore, if we want to make a link look like a button, then it should be achieved either by providing custom CSS or by having a separate component, i.e. `LinkButton`. The first approach is already supported in the current library, so I focused on the second one.

With the changes from this PR, two new components are introduced and one is simplified:
* `AnchorButton` — renders `<a />` that looks like a Gatsby button. Uses `BaseAnchor`, a new component, under the hood.
* `LinkButton` — renders Gatsby `<Link />` that looks like a Gatsby button. Uses `BaseLink`, a new component, under the hood.
* `Button` — doesn't do any conditional tag/component rendering anymore; now it always renders `<button />`. Uses `BaseButton` under the hood.

### Declaration files
`Button`, which is the main focus of TypeScript integration for in PR, uses some code that is not TS yet. To avoid bloating the task with rewriting all of that right now, we can use declaration files to just describe what are the types for that code. For example, [here](https://github.com/gatsby-inc/gatsby-interface/compare/copy-button...ts-button-poc?expand=1#diff-7ff0ac5d650116e059b2cf425055c99a) we declare type for the default export of `colors`.

### Replacing prop types
With TypeScript, declaring and using component prop types becomes more clean and detailed:
* Engineers **have to** provide prop types and keep them up-to-date, otherwise the code just won't compile. Granted, it is still possible to just declare some prop as `any`, but that is easy to spot during code review. Make sure to export the props type so that it can be used by other components. 
* More detailed and helpful types for non-primitive props — a prop that accepts one of predefined values can use a union or an enum type (see `ButtonVariant`, `ButtonSize` and `ButtonTone` in this PR`); the type itself can be easily exported for other components to use.
* Reduce ambiguity — one of the reasons to split `BaseButton` (and `Button`) into 3 components is that TypeScript does not handle conditional props very well. In some cases this might become a nuisance (though I've only experienced this once or twice before), in the case of `Button` it actually makes sense (see **Split responsibilities** above).
* Easily extend prop types — since `Button` is just a wrapper for `BaseButton`, it can serve as a great example for how to work with prop types in TypeScript. `BaseButton` exposes its prop types, `BaseButtonProps`, for `ButtonProps` to extend:
  ```typescript
  import { BaseButton, BaseButtonProps } from "../../skeletons/BaseButton"
  
  export type ButtonStyleProps = {
    size?: ButtonSize
    tone?: ButtonTone
    variant?: ButtonVariant
  }
  
  export type ButtonProps = BaseButtonProps & ButtonStyleProps
  ```
  At the same time, `BaseButtonProps` are just extending types for native HTML `<button />`:
  ```typescript
    export type BaseButtonProps = Omit<JSX.IntrinsicElements["button"], "ref"> &
      BaseElementContentProps & {
        loading?: boolean
        loadingLabel?: React.ReactNode
        LoadingIcon?: React.ComponentType<any>
    }
  ```
  If your type for component props extends `JSX.IntrinsicElements`, make sure to omit `ref` prop — it is defined for HTML elements by default, but it is not available for our custom components unless we create them using `React.forwardRef`.

### Forwarding refs
Generally it makes sense to forward refs for any component which is just a wrapper for an interactive HTML element (e.g. `button`, `input`, `a`). This will allow library users to easily implement some of the common features for those elements, like focusing an input on mount.

With TypeScript, `React.forwardRef` is a function with two generic types: one for the referenced value (e.g. `HTMLAnchorElement` if the ref is passed to an `<a />` element) and the component prop types.

This is how `BaseAnchor` is defined to forward ref:
```typescript

export type BaseAnchorProps = Omit<JSX.IntrinsicElements["a"], "ref"> &
  BaseElementContentProps & {}

const BaseAnchor = React.forwardRef<HTMLAnchorElement, BaseAnchorProps>((props, ref) => {
  /* render anchor */
})
```

## My Questions
Most of these are just related to implementation. 

- [x] Why do we have a `DefaultIcon` prop for `BaseButton`? Is it just for the `label` shortcut (which is also related to my next question)? 
- [x] Do we really need the `label` prop for `Button`? It adds ambiguity to how this component can be used. Surely writing `<Button label="Click me" DefaultIcon={MdIconCheck} />` isn't much more optimal than `<Button>Click me <MdIconCheck /></Button>`?
- [ ] Why do we wrap string children of `Content` (`BaseElementContent` in this PR) in `<span />`? I feel like if there's a specific reason for this, we should add it as a comment.


greglobinski:
> Why do we have a DefaultIcon prop for BaseButton? Is it just for the label shortcut (which is also related to my next question)?

The intention was the shortcut usage

`<Button label="This is Label" />`

which should produce the 'BRAND' tone, 'PRIMARY' variant button with 'MdArrowForward' icon, because seemed to me as a most often use case.

But seems that it's not so useful as I thought then, so +1 to remove it.

> Do we really need the label prop for Button? It adds ambiguity to how this component can be used.

Based on the previous one, the answer is "No"

> Why do we wrap string children of Content (BaseElementContent in this PR) in <span />? I feel like if there's a specific reason for this, we should add it as a comment.

That is connected with  styling/positioning svg-icons, I have to dig into it, and will add a comment or ask for suggestions.

Elanhant:
> That is connected with styling/positioning svg-icons, I have to dig into it, and will add a comment or ask for suggestions.

I wonder if we could use flexbox for this by making button inline-flex with center align?

Elanhant:
@greglobinski sounds good! What do you think about passing icons themselves as props? Something like
```jsx
<Button leftIcon={<MdSync />} rightIcon={<MdArrowForward />}>
  Click me!
</Button>
```
In this case people using the component won't have to remember to provide boolean `iconOnLeft` and `iconOnRight` props whenever they are using an icon.

greglobinski:
> people using the component won't have to remember to provide boolean

@Elanhant I like it

Elanhant:
@greglobinski awesome! If you don't mind, I can handle that

greglobinski:
@Elanhant sure

greglobinski:
@Elanhant I've committed a tiny fix for positioning `loading` icon

PR #70, opened 2019-10-15
(WIP): Port dotcom navigation to gatsby-interface

amberleyromo:
## Description

This PR ports the dotcom navigation to the gatsby-interface package.

## Notes

- New component `<Navigation>`, with subcomponents:
	- `<Dropdown>`
	- `<Item>`
- New shared utils
	- Ported the navigation helpers into `src/utils/helpers/navigation.js`
- New presets (verbatim from mansion)
	- `src/utils/presets/media-queries.js`
	- `src/utils/presets/shadows.js`
	- `src/utils/presets/transition.js`
- Modified presets
	- `src/utils/presets/colors.js`
		- Added gray variations to "gray" key in palette obj.
	- `src/utils/presets/sizes.js`
		- Added new `radius` option `small`
		- Added `siteHeader` and `layout` options to `dimensions`
	- `src/utils/presets/spaces.js`
		- Added a "2m" spacing option between "m" and "l".
- Added stories for:
	- default
	- inverse

### Default

Screen Shot 2019-10-15 at 1 11 51 PM

### Inverted

Screen Shot 2019-10-15 at 1 12 04 PM

## Questions

- [ ] I moved over `src/modules/env/constants.js`. Is this something we want in this package, or should be handled differently?
- [ ] Agree with the modification to `src/utils/presets/spaces.js`? Bit hinky but I wanted to preserve the existing structure.
- [ ] In terms of file structure, I kept everything together (like the `<Card>` component) but for now I'm only exporting the main `<Navigation>` component (not `<Dropdown>` or `<NavItem>`. I think that makes sense for now, and we can abstract it further when we have a use for that. For now, it's all still tightly coupled. Thoughts?
- [ ] Similar question for handling of nav items. For now, the helper function `getNavItems` expects an array of navigation items consistent with how we're sourcing from Contentful. Should these methods not be part of gatsby-interface, and instead supplied where the component is imported and used, and document the structure the component expects? 

## Related issues

gatsby-inc/mansion#4593


rizchelwood:
Can you change the base branch to `dev` instead of `master`?

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+  radius,
+  transition,
+  colors,
+  palette,
+  fontSizes,
+  shadows,
+  spaces,
+} from "../../utils/presets"
+
+export const Dropdown = ({ children, bpToggleNavigation }) => (
```
rizchelwood: Let's do these as compound components - i.e. `Dropdown` and `Dropdown.Item`. You can look at `SidebarNav` as an example. https://kentcdodds.com/blog/compound-components-with-react-hooks

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+  <ul
+    className="child"
+    css={{
+      [bpToggleNavigation]: {
```
rizchelwood: Is there a better name for this? It's not really clear what it is

src/components/Navigation/Dropdown.js
rizchelwood: You can also provide a shortcut like we do in the other compound components we use. If they pass in `items` (or whatever name prop), you can render the dropdown items and use only the `Dropdown` component. Kind of like this https://github.com/gatsby-inc/gatsby-interface/blob/master/src/components/core/IntegrationRow/IntegrationRow.js

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+
+const navItemStyles = {
+  textDecoration: `none`,
+  color: colors.white,
+  fontFamily: fontFamilies.headerFontFamily,
+  fontWeight: `normal`,
+  transition: `opacity ${transition.speed.default}`,
+  WebkitFontSmoothing: `antialiased`,
+  fontSize: fontSizes[`2xl`],
+  "&.nav-item-active": {
```
rizchelwood: Instead of making these dependent on `classNames`, we should use this as a variant `ACTIVE`

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+)
+
+export const DropdownItem = ({ children, bpToggleNavigation }) => (
+  <li
+    css={{
+      margin: 0,
+      position: `relative`,
+      zIndex: 1,
+      listStyle: `none`,
+      "& .nav-link-sub": {
```
rizchelwood: Same here. Let's not depend on `classNames` and use a prop. Or create another compound component and apply these styles (whichever makes more sense)

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+    color: colors.accent,
+    "&.is-inverted": {
```
rizchelwood: Same here. We could have an tone prop `INVERTED`

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+  colors,
+  palette,
+  fontSizes,
+  shadows,
+  spaces,
+} from "../../utils/presets"
+
+export const Dropdown = ({ children, bpToggleNavigation }) => (
+  <ul
+    className="child"
```
Elanhant: This looks like a magic string, and I'm also not sure if the only usage of this in the PR is actually happening. Can we verify if it works?

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+        background: `linear-gradient(45deg, ${colors.accent}, ${colors.accent})`,
+      },
+    },
+  },
+  "&:hover": {
+    opacity: 0.8,
+  },
+}
+
+const NavItem = ({
```
rizchelwood: You can make this a compound component of `Navigation`

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+    css={{
```
Elanhant: I'd suggest moving huge style definitions like this to a separate const outside the component body. Otherwise it is getting really hard to see what is the actual JSX :)

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+  sections,
+  isInverted,
```
rizchelwood: Like comment above - you can make this into a tone prop

src/components/Navigation/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+      },
+    }}
+  >
+    {children}
+  </ul>
+)
+
+export const DropdownItem = ({ children, bpToggleNavigation }) => (
+  <li
+    css={{
```
Elanhant: Ditto for extracting big chunks of CSS

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+  linkTo,
+  children,
+  isLast,
+  bpToggleNavigation,
+  updateParentState,
+}) => {
+  const [isMobileNavOpen] = useState(false)
```
Elanhant: This looks weird. Why `useState` if it's not actually updated?

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+          position: `relative`,
+          WebkitFontSmoothing: isInverted ? `antialiased` : false,
+          ":hover .child": {
+            display: `inline-block`,
+          },
+        },
+      }}
+    >
+      {isLast &&
+      (linkTo.includes(`start-project`) || linkTo.includes(`get-started`)) ? (
```
rizchelwood: We should make this more generic so it can be applied anywhere. Also last element might not always be a button (it wasn't before) so this component should be able to have a button or list item as a last component or in any order (first item, etc). 

I think making this component more dynamic in general would be helpful.

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+              color: isInverted ? palette.purple[600] : palette.white,
+            },
+          }}
+        >
+          {children}
+        </Button>
+      ) : (
+        <Link
+          to={linkTo}
+          css={{
```
Elanhant: Ditto for extracting big chunks of CSS

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+      {isLast &&
+      (linkTo.includes(`start-project`) || linkTo.includes(`get-started`)) ? (
+        <Button
+          to={linkTo}
+          size={`M`}
+          css={{
+            background: isInverted ? palette.white : colors.gatsby,
+            border: isInverted ? `1px solid ${palette.purple[200]}` : `0`,
+            color: isInverted ? palette.purple[500] : palette.white,
+            fontSize: `1rem`,
```
rizchelwood: Pull this from the `fontSizes` presets

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+              fontFamily: fontFamilies.bodyFontFamily,
+              fontWeight: `normal`,
+              color: `inherit`,
+              textDecoration: `none`,
+              lineHeight: `calc(2 * ${dimensions.siteHeader.heightLarge} / 3)`,
+              position: `relative`,
+              "& span": {
+                position: `relative`,
+              },
+              "&.nav-item-active": {
```
rizchelwood: Same as comment above

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+            },
+          }}
+          className={isInverted ? `nav-item is-inverted` : `nav-item`}
+          activeClassName="nav-item-active"
+          onClick={() => updateParentState(!isMobileNavOpen)}
+        >
+          <span>{children}</span>
+        </Link>
+      )}
+      {sections && sections.length > 0 && (
```
rizchelwood: Is there a better name for `sections`? It's not very clear what it is

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  render() {
+    const { isInverted, bpToggleNavigation } = this.props
+    const navItems = getNavItems(this.props.navItems)
+
+    return (
+      <Fragment>
+        <Global
+          styles={css`
```
Elanhant: Ditto for extracting big chunks of CSS

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,207 @@
...
+
+import NavItem from "./Item"
+import { getNavItems, getLink } from "../../utils/helpers"
+import { dimensions, palette, spaces, mediaQueries } from "../../utils/presets"
+import {
+  newLandingPagesEnabled,
+  newOnboardingEnabled,
+} from "../../modules/env/constants"
+
+class Navigation extends React.Component {
```
Elanhant: We should probably rewrite this as a functional component (especially since there is an `UNSAFE_` method here)

src/components/Navigation/Item.js
rizchelwood: Just an idea but you could make a `Navigation` compound component for `Button` and `Link`. It would make the component more dynamic/composable and clean a lot of this up

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,207 @@
...
+              },
+            }}
+          >
+            {navItems.map((navItem, i) => {
+              if (i + 1 >= navItems.length) {
+                let linkTo
+                let anchorText
+                let icon
+
+                if (newLandingPagesEnabled) {
```
Elanhant: This should be delegated to component consumers

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,168 @@
...
+        <Dropdown bpToggleNavigation={bpToggleNavigation}>
+          {sections.map(section => (
+            <DropdownItem
+              key={`subnav-${section.name}`}
+              bpToggleNavigation={bpToggleNavigation}
+            >
+              <Link to={getLink(section)} className="nav-link-sub">
```
rizchelwood: You can pass `SUB` as a variant instead of `className` and apply styles that way

src/components/Navigation/Navigation.js
rizchelwood: Can you make this into a functional component and using react hooks?

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,207 @@
...
+    })
+  }
+
+  render() {
+    const { isInverted, bpToggleNavigation } = this.props
+    const navItems = getNavItems(this.props.navItems)
+
+    return (
+      <Fragment>
+        <Global
```
rizchelwood: This seems a bit hacky?

src/utils/presets/spaces.js
```diff
@@ -4,6 +4,7 @@ export const spaces = {
   xs: `0.5rem`,
   s: `0.75rem`,
   m: `1rem`,
+  "2m": `1.25rem`,
```
Elanhant: Is it possible to just use `m`? I think we should be extremely cautious when it comes to adding new values to presets, otherwise those can get bloated really easy and the UI won't look consistent.

src/modules/env/constants.js
```diff
@@ -0,0 +1,7 @@
+export const isGA = process.env.GATSBY_GA === `true`
```
rizchelwood: I don't think we should be bringing over the env vars from cloud. Maybe we could pass this in as a prop or just handle this in .com?

src/components/Navigation/Dropdown.js
amberleyromo: Sure, I'll think on that. It's the breakpoint at which the navigation style breaks into hamburger/mobile style.

rizchelwood: Maybe it's just `bp` that's throwing me off. `toggleNavigation` makes sense

Elanhant: `toggleNavigation` sounds rather like a function than a scalar. I think something like `mobileViewMedia` would be better. Or we could go in a different direction and introduce a boolean `isMobileView` prop and apply mobile styles only when it's true.

src/components/Navigation/Item.js
amberleyromo: "sections" are a specific content type from Contentful -- hence my question about how tightly coupled this navigation is to our Contentful model, and the relevance to how we build for gatsby-interface.

src/components/Navigation/Dropdown.js
amberleyromo: Can you clarify "magic string"? Regardless, looks like it's used for the dropdown item hover display, definitely a better way to do that.

src/components/Navigation/Item.js
rizchelwood: We shouldn't have this tightly coupled to Contentful. We want it to be able to work with any kind of data source. What I would suggest is passing in a prop to this component that can come from any data source (i.e. for this example, pass in `section` from .com to the component with a more generic prop name). This way you're just passing in the data in as a prop and it can come from anywhere

src/components/Navigation/Dropdown.js
Elanhant: "Magic" in a sense that we have a string `"child"` which comes from nowhere and is very generic; this makes it harder to work with it (e.g. if we need to refactor the code that uses this class name or just search for occurrences)

src/utils/presets/spaces.js
amberleyromo: It's certainly possible to just use `m`, it just deviates from the current styling a touch. Agree about caution. Happy to remove.

src/modules/env/constants.js
amberleyromo: I think handling in dotcom makes sense for these, a la https://github.com/gatsby-inc/gatsby-interface/pull/70#discussion_r335134414

src/components/Navigation/Dropdown.js
amberleyromo: Couldn't agree more! I still have some refactoring to do for this legacy code from `mansion` codebase. I was more opening this up for the other discussion points, but all comments appreciated!

src/components/Navigation/Item.js
amberleyromo: It does look like it's used, it's just initialized here as false. I haven't dug much into this in particular though, I'll take a closer look 👍🏼

src/components/Navigation/Item.js
amberleyromo: So this is styling the nav item for the active page. (For example, if you're on the "About" page, the "About" link will be styled differently, indicating it's the active page). 

(If you already know this, sorry, I actually didn't, so I'm prefacing!) [`activeClassName` is built into Gatsby Link](https://www.gatsbyjs.org/docs/gatsby-link/#add-custom-styles-for-the-currently-active-link) providing a hatch to style the active link.

I'm not certain of the best approach to use a tone prop for this, but it would include reimplementing logic to check the active link.

1) Do we want to do that, or take advantage of `activeClassName`?
2) Do you have a different idea I haven't thought of?

src/components/Navigation/Dropdown.js
amberleyromo: Refactored to compound components, and changed the dropdown to pass in `items`.

src/components/Navigation/Dropdown.js
amberleyromo: For now I've left it intact as a media query, and renamed it `mobileNavMediaQuery`

src/components/Navigation/Item.js
greglobinski: I think that if we can, we can stay with Gatsby Link built-in functionality

src/components/Navigation/Item.js
greglobinski: > We shouldn't have this tightly coupled to Contentful.

:100:

PR #73, opened 2019-10-18
Add font info to README

joshwcomeau:
Storybook requires a few fonts to run, but they're `.gitignore`d, since they're licensed. I uploaded those fonts to our (private) Google Drive, and added a blurb to the README about where to find them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@
 yarn storybook
 ```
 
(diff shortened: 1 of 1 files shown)


rizchelwood:
Can you change the base branch to `dev`? Other than that, LGTM! Thanks for doing this

PR #77, opened 2019-10-18
Typescript Guidelines

Elanhant:
Just some general suggestions on how to write React components in TypeScript based on my prior experience with core components libraries. The document content is presented below for convenience.

---- 
# TypeScript guidelines

This document's goal is to provide some guidelines and suggestions on using TypeScript when working on components in `gatsby-interface`.

##### Table of Contents

- [Prop Types](#prop-types)
  - [Naming](#naming)
  - [Optional props](#optional-props)
  - [Always export prop types](#always-export-prop-types)
  - [ReactNode](#reactnode)
  - [Be strict](#be-strict)
  - [Mirror HTML attributes](#mirror-html-attributes)
  - [Avoid conditional prop types](#avoid-conditional-prop-types)
- [React Hooks](#react-hooks)
  - [useState](#usestate)
  - [useCallback](#usecallback)
  - [useRef](#useref)
- [React Types](#react-types)
  - [React.FC](#reactfc)
  - [React.forwardRef](#reactforwardRef)
  - [Event handlers](#event-handlers)
  - [Context](#context)
- [Emotion](#emotion)
- [Typing Third-party Libraries](#typing-third-party-libraries)

## Prop Types

TypeScript provides an alternative to [prop-types](https://github.com/facebook/prop-types) library when it comes to declaring types for a component props.

There are several differences between the two:

- Type checks with TypeScript are performed during **compilation**, while `prop-types` performs runtime validation
- TypeScript **forces** developers to provide types to their components props, otherwise the code just won't compile
- TypeScript provides better support for typing functions, while with `prop-types` we can only use `PropTypes.func` type
- TypeScript provides better ecosystem for types: we can extend them, use a partial type, declare a union type, export a type, use generics and enums etc

Here are some general rules for writing a type for component props:

### Naming

A TypeScript type for a component should be name `[componentName]Props`, e.g. `MyComponent` should have props named `MyComponentProps`.

### Optional props

If some prop is not considered required for your component, you can use `?` to [mark it as optional](https://www.typescriptlang.org/docs/handbook/interfaces.html#optional-properties) in prop types. This approach is different to `prop-types` library where you mark props as required.

```typescript
type MyDateSelectorProps = {
  value: string // required
  onChange: (value: string) => void // required
  mode?: "input" | "widget" // optional, same as "input" | "widget" | undefined
}
```

### Always export prop types

1. If a component is exported, the type should be exported as well. This allows other components to use this type in their own prop types. If the type uses some enum types, it usually makes sense to export them as well.

```tsx
// MyComponentProps and MyComponentVariant types are exported

export type MyComponentProps = {
  content: string
  variant: MyComponentVariant
}

export type MyComponentVariant = "foo" | "bar"

export default function MyComponent(props: MyComponentProps) {
  /* ... */
}
```

### ReactNode

`React.ReactNode` is a React type which covers everything that can be used as a `children` prop. Therefore, it's better to use this type for any kind of prop that gets rendered as a child in an HTML node and resort to strings only if it can or must be used as an attribute. For example:

```tsx
export type PageTitleProps = {
  title: React.ReactNode
  subtitle?: React.ReactNode
}

export default function PageTitle({ title, subtitle }) {
  return (
    <h1>
      {title}
      {subtitle && <small>{subtitle}</small>}
    </h1>
  )
}
```

The reason for this is components can be used not only to style and/or compose elements, but also to render a transformed value. Take [react-intl](https://github.com/formatjs/react-intl/), for example: it exposes a [FormattedMessage](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage) component which is used to display translated text. If our example above only allowed `title` and `subtitle` as strings, a consumer application would not be able to use `PageTitle` like this:

```jsx
import { PageTitle } from "gatsby-interface"
import { FormattedMessage } from "react-intl"

export default function HomePageTitle({ title, subtitle }) {
  return (
    <h1>
      <PageTitle
        title={<FormattedMessage id="homePage.title" />}
        subtitle={<FormattedMessage id="homePage.subtitle" />}
      />
    </h1>
  )
}
```

### Be strict

Try to be as strict as possible when declaring types for your props. Avoid using `any` or `unknown` types as well as `Function` or `object` anywhere but development-related code (i.e. tests, stories and WIP code). If your component renders a button which calls `handleButtonClick` prop when clicked, then it should have either of these type declarations:

```tsx
export type MyComponentWithButtonProps = {
  handleButtonClick: () => {} // implies that the consumer component does not care about the click event
}
```

or

```tsx
export type MyComponentWithButtonProps = {
  handleButtonClick: React.MouseEventHandler<HTMLButtonElement> // allows the consumer component to access the event
}
```

### Mirror HTML attributes

If your component is just a wrapper for a HTML element, its prop types should be based on the attributes supported by that element. This can be achieved with `JSX.IntrinsicElements` type:

```tsx
export type MyLogButtonProps = Omit<JSX.IntrinsicElements["button"], "ref">

export default function MyLogButton({ onClick, ...rest }: MyLogButtonProps) {
  return (
    <button
      onClick={e => {
        console.log(e)
        onClick && onClick(e)
      }}
      {...rest}
    />
  )
}
```

**NOTE:** remember to omit `ref` from JSX.IntrinsicElements since your component does not actually handle refs.

### Avoid conditional prop types

In some cases it is tempting to create a component that renders different components with different props based on some subset of its own props. Naturally, a component like this will have to either calculate relevant props itself or allow passing them from a parent component. Here are examples to illustrate both cases:

```tsx
type MyDateSelectorProps = {
  value: string
  onChange: (value: string) => void
  mode?: "input" | "widget"
}

// Passed "mode" prop is used to provide different "onChange" props for <input /> and <DatePickerWidget />
function MyDateSelector({
  value,
  onChange,
  mode = "widget",
}: MyDateSelectorProps) {
  if (mode === "input") {
    return (
      <input
        type="date"
        value={value}
        onChange={e => onChange(e.target.value)}
      />
    )
  }
  return <DatePickerWidget value={value} onChange={onChange} />
}
```

and

```tsx
type MyButtonLookalikeProps = Omit<
  JSX.IntrinsicElements["a"] & JSX.IntrinsicElements["button"],
  "ref"
>

// The component just passes ...rest to <a /> and <button />
function MyButtonLookalike({ href, ...rest }: MyButtonLookalikeProps) {
  if (href) {
    return <a href={href} {...rest} className="my-btn" />
  }
  return <button {...rest} className="my-btn" />
}
```

The second case won't even compile since `rest` may contain attributes not recognized by `<button />` such as `rel` and `target`.
It is possible to have conditional prop types with TypeScript; however, it usually makes more sense to see if the component actually needs that. In our example above, a `MyButtonLookalike` is used to render either an anchor link or a button, and both of them should look like a button. Yet this makes the component more ambiguous since we'd have to look at the props to see what will be actually rendered:

```jsx
// In this case we can't even know what gets rendered - is it always a button? an anchor?
function TextWithButton({ text, ...rest }: MyButtonLookalikeProps & { text: string }) {
    return (
        <p>
            {text}
            <MyButtonLookalike {...rest}>
        </p>
    )
}
```

Instead, we can split this component in two and extract the shared styles logic:

```tsx
function getButtonStyles(): { className: string } {
  return {
    className: "my-btn",
  }
}

type MyButtonProps = Omit<JSX.IntrinsicElements["button"], "ref">

function MyButton(props: MyButtonProps) {
  return <button {...props} {...getButtonStyles()} />
}

type MyAnchorButtonProps = Omit<JSX.IntrinsicElements["a"], "ref">

function MyAnchorButton(props: MyAnchorButtonProps) {
  return <a {...props} {...getButtonStyles()} />
}
```

### useState

```typescript
/*
 * useState can infer types for "step" and "setStep" from the default value type
 * However, an explicit declaration makes the code more clear to other contributors
 */
const [step, setStep] = React.useState<number>(1)
```

### useCallback

```typescript
/*
 * Declare types the same way you would do it for an arrow function
 */
const onRemove = React.useCallback((itemId: string) => {
  /* ... */
}, [])

// or, if there is a type alias for this type of functions:
type RemoveHandler = (itemId: string) => void

const onRemove = React.useCallback<RemoveHandler>(itemId => {
  /* ... */
}, [])

// event handlers
const onFocus = React.useCallback<React.FocusEventHandler<HTMLInputElement>>(
  e => {
    /* ... */
  },
  []
)
```

### useRef

- The type that you provide will be used for `ref.current`, not for the `ref` itself!
- Passing `null` as default value to `useRef` makes TypeScript treat the ref as **readonly**

```tsx
const ref = React.useRef<HTMLButtonElement>(null)

return <button ref={ref}>Click me!</button>
```

### React.FC

If you're a fan of declaring components as arrow functions, you can use a shortcut `React.FC<T>` &mdash; it injects `children: React.ReactNode` into the type `T` that you provide and provides a return type. So, instead of writing

```typescript
export type MyComponentProps = {
  children?: React.ReactNode
  someRandomNumber: number
}

function MyComponent({ children, someRandomNumber }: MyComponentProps) {
  /* ... */
}
```

you could have

```typescript
export type MyComponentProps = {
    someRandomNumber: number;
}

const MyComponent: React.FC<MyComponentProps> = ({ children, someRandomNumber }: MyComponentProps) {
    /* ... */
}

export default MyComponent
```

However, this can mess up prop types for components that import `MyComponentProps` since they might expect that `children` would be there which isn't true for the second case.

### React.forwardRef

`React.forwardRef` accepts two generic types: ref type and prop types:

```typescript
export type MyToggleProps = {
  id: string
  label: React.ReactNode
  onToggle: (value: boolean) => void
}

const MyToggle = React.forwardRef<HTMLInputElement, MyToggleProps>(
  ({ id, label, onToggle }, ref) => {
    return (
      <React.Fragment>
        <input
          id={id}
          ref={ref}
          type="checkbox"
          onChange={e => onToggle(!e.target.checked)}
        />
        <label htmlFor={id}>{label}</label>
      </React.Fragment>
    )
  }
)

export default MyToggle
```

### Event handlers

React has type aliases for basically all `onX` props that can be used on native HTML elements. Here are some examples (generic type `<T>` usually refers to the element that the event is attached to, i.e. `e.currentTarget`):

- `onClick`, `onMouseEnter`, `onMouseLeave`, `onMouseDown` etc &mdash; `React.MouseEventHandler<T>`
- `onFocus`, `onBlur` &mdash; `React.FocusEventHandler<T>`
- `onChange` for `input`, `select` and `textarea` &mdash; `React.ChangeEventHandler<T>`
- `onSubmit` &mdash; `React.FormEventHandler<T>`

You can use these aliases in your prop types for props.

### Context

Typing React context is pretty straightforward: `React.createContext` is a generic function in TypeScript that accepts a type for the context value. Just provide that type and it will be inferred by context provider and consumers. For example, this is how a `ToastContext` would be defined in TS:

```typescript
// Type for the context value
export type ToastContextValue = {
  showToast: (message: React.ReactNode, options?: ToastOptions) => void
}

export type ToastOptions = {
  tone?: "SUCCESS" | "DANGER"
  timeout?: number
}

// Pass ToastContextValue to generic React.createContext
export const ToastContext = React.createContext<ToastContextValue>({
  showToast: () => {}, // default context value must respect the type
})

/*
 * TypeScript will infer the props types ("value") for ToastContext.Provider from ToastContext
 */
export const ToastProvider = ToastContext.Provider

/*
 * TypeScript will infer the type for ToastContext.Consumer render function from ToastContext
 */
export const ToastConsumer = ToastContext.Consumer

/*
 * TypeScript will infer the return type for from React.useContext
 * which in turn infers it from ToastContext
 */
export function useToast() {
  return React.useContext(ToastContext)
}
```

## Emotion

It is possible to provide prop types for your `styled` components:

```typescript
export const MorpheusPill = styled("div")<{ isMatrixPill: boolean }>`
  color: ${props => (props.isMatrixPill ? "red" : "blue")};
`
```

However, it might be better to extract those types in the same way as we do for other prop types:

```typescript
export type MorpheusPillProps = {
  isMatrixPill: boolean
}

export const MorpheusPill = styled("div")<MorpheusPillProps>`
  color: ${props => (props.isMatrixPill ? "red" : "blue")};
`
```

This approach is also helpful if you're using VSCode and [vscode-styled-components](https://marketplace.visualstudio.com/items?itemName=jpoissonnier.vscode-styled-components) plugin since it only works if the template string starts on the same line as the `styled`/`css` call.
If we would leave the type declaration in the generic, our code might be prettified into something like this:

```typescript
export const MorpheusPill = styled("div")<{
  isMatrixPill: boolean // this would break vscode-styled-components
}>`
  color: ${props => (props.isMatrixPill ? "red" : "blue")};
`
```

## Typing Third-party Libraries

Many of JavaScript libraries out there do provide typings either as part of the package itself or available as [@types/library-name](https://github.com/DefinitelyTyped/DefinitelyTyped). However, some libraries still have no types exposed. The lack of type definitions for third-party libraries must never stop us from using those libraries, but luckily there is a workaround for this: just add a broad type definition in the [/src/typings.d.ts](/src/typings.d.ts) file, e.g.:

```typescript
declare module "@xstyled/system"
```

If the library is used extensively in our code then we could try to provide our own typings if it is not too difficult or time-consuming. For example:

```typescript
declare module "@xstyled/styled-components" {
  import styled, {
    css as styledCss,
    createGlobalStyle as styledCreateGlobalStyle,
  } from "styled-components"

  export default styled
  export const css: typeof styledCss
  export const createGlobalStyle: typeof styledCreateGlobalStyle
}
```


docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+const ref = React.useRef<HTMLButtonElement>(null)
+
+return <button ref={ref}>Click me!</button>
+```
+
+## React types
+
+### React.FC
+
+If you're a fan of declaring components as arrow functions, you can use a shortcut `React.FC<T>` &mdash; it injects `children: React.ReactNode` into the type `T` that you provide. So, instead of writing
```
joshwcomeau: Huh, is that all it does? I assumed it also manages the return type, expecting it to be a React element or `null`

mfrachet: @joshwcomeau I think you're right, it provides the return type `React.ReactElement` that can be null 😊

docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+
+```typescript
+export type MyComponentProps = {
+  children?: React.ReactNode
```
mfrachet: It could be interesting to detail about the optional parameters / props wyt?

docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+export type MyComponentVariant = "foo" | "bar"
+
+export default function MyComponent(props: MyComponentProps) {
+  /* ... */
+}
+```
+
+### ReactNode
+
+`React.ReactNode` is a React type which covers everything that can be used as a `children` prop. Therefore, it's better to use this type for any kind of prop that gets rendered as a child in an HTML node and resort to strings only if it can or must be used as an attribute. For example:
```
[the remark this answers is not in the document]
Elanhant: I can see why we'd want to restrict props to string if we want to enforce some styling, however there is a counterargument that I've found while working with those `string` props — we might want to pass component that does something completely unrelated to styles. One of the examples would be `FormattedMessage` from [react-intl](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage). Forcing `string` to props in my example would not allow consumer applications to use that component as intended:
```jsx
import { PageTitle } from 'gatsby-interface'
import { FormattedMessage } from 'react-intl'

export default function HomePageTitle({ title, subtitle }) {
  return (
    <h1>
      <PageTitle
        title={<FormattedMessage id="homePage.title" />}
        subtitle={<FormattedMessage id="homePage.subtitle" />}
      />
    </h1>
  )
}
```

docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+      {title}
+      {subtitle && <small>{subtitle}</small>}
+    </h1>
+  )
+}
+```
+
+### Be strict
+
+Try to be as strict as possible when declaring types for your props. Avoid using `any` or `unknown` types as well as `Function` or `object`. If your component renders a button which calls `handleButtonClick` prop when clicked, then it should have either of these type declarations:
```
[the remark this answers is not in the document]
Elanhant: Well I've been working on a core components library with TypeScript with about 35 components, and `any` is used there only as a workaround for issues with typings for third-party libraries and in some tests. Can't say it was too much of a hassle or inconvenience to achieve that, and I'm ready to help out people with finding a proper type for their props :)

I'd say it's fine to use `any` or `unknown` while working on something new, but I feel strongly against having `any` in master unless accompanied by a comment explaining why it is there.

docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+    /* ... */
+  },
+  []
+)
+```
+
+### useRef
+
+- The type that you provide will be used for `ref.current`, not for the `ref` itself!
+- Passing `null` as default value to `useRef` makes TypeScript treat the ref as **readonly**
```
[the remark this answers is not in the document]
Elanhant: Hm, not sure why it doesn't repro in sandbox, but it does repro locally in `gatsby-interface` when I change `CopyButton`:
```typescript
  // added "null" as default value
  const timeoutRef = React.useRef<number | undefined>(null)

  const copyOnClick: React.MouseEventHandler<HTMLButtonElement> = async e => {
    /* ... */
    // TypeScript complains with "Cannot assign to 'current' because it is a read-only property"
    timeoutRef.current = window.setTimeout(() => {
      setCopied(false)
    }, delay)
  }
```

Also, check [this issue thread](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31065#issuecomment-453841404). It's still open so maybe they'll change this behaviour, but for now TS is supposed to complain about such cases.

docs/typescript.md
Elanhant: Ah yes, forgot about return types since TypeScript usually does a good job on inferring them 🙂

Elanhant:
Thanks for the feedback, @joshwcomeau and @mfrachet!

Elanhant:
> * How to type React context (selfishly I want this because this keeps biting me)
Sure, here's an example, and I'll add a section to the doc as well later:
```typescript
// ToastProvider.tsx
import React from 'react'

export type ToastOptions = {
  tone?: 'SUCCESS' | 'DANGER';
  timeout?: number;
}

export type ToastContextValue = {
  showToast: (message: React.ReactNode, options?: ToastOptions) => void
}

export const ToastContext = React.createContext<ToastContextValue>({
  showToast: () => {},
})
```
That's basically it; `ToastContext.Provider`, `ToastContext.Consumer` and `useContext(ToastContext)` will all infer `ToastContextValue` for their types.

Elanhant:
> * Typing `styled.whatever` with Emotion

`styled` accepts a generic type which allows you to provide expected prop types: 
You can type a styled component like this:
```typescript
export const MorpheusPill = styled('div')<{ isMatrixPill: boolean }>`
  color: ${props => props.isMatrixPill ? 'red' : 'blue'};
`
```

Elanhant:
> * How to deal with 3rd-party deps that don't make type definitions available (my sentiment here is that it isn't worth the trouble, and such dependencies are getting rarer and rarer, but curious to know if you have different thoughts)

The lack of type defs for third-party libraries must never stop us from using those libraries, so I'm totally fine with just adding a broad type def in our [srs/typings.d.ts file](https://github.com/gatsby-inc/gatsby-interface/blob/master/src/typings.d.ts), smth like:
```typescript
declare module "@xstyled/system"
```

docs/typescript.md
Elanhant: I think it's more relevant to the area of general TS knowledge sine this is really related to typing an object, though it might be useful to add a small note on this when used with props. Will update the doc 👍

greglobinski:
Thank you @Elanhant, that's really helpful, a lot of new stuff for me. Not now, but I guess I will have questions pretty soon :). I plan to start with rewriting one of the existing component to TS, that should give me more practical understanding about TS.

docs/typescript.md
```diff
@@ -0,0 +1,324 @@
...
+function MyButton(props: MyButtonProps) {
+  return <button {...props} {...getButtonStyles()} />
+}
+
+type MyAnchorButtonProps = Omit<JSX.IntrinsicElements["a"], "ref">
+
+function MyAnchorButton(props: MyAnchorButtonProps) {
+  return <a {...props} {...getButtonStyles()} />
+}
+```
```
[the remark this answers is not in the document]
Elanhant: Well the issue with this particular example (and with the `Button` component from `gatsby-interface` before TS was introduces) is that the component itself is too vague to use. Let me try to present my point of view in a more clear way.

The `Button` component used to represent three distinct functionalities: a button, an anchor tag, a "gatsby" link. This thing alone made it confusing to use the component as it conflates three different APIs into one on the basis of styling. Imagine if we had a component named `FormControl` that would render either an `<input />` (with any of the supported `type` attribute values) or a `<select />` or a `<textarea />` or even a date picker, all of them sharing the same style. 
The only advantage of this approach I can think of is having everything in one place; yet this component can very quickly became quite a beast to maintain if we want to add some new prop/play with the styles etc. Not to mention that it's going to be confusing to the developers since some props might be the same and some might not and they will have to figure out what props they can pass. 

Another point is that for a "core" components library I would expect components like `Button` to mimic their HTML counterparts; and we don't use a `<button />` for links, do we? 🙂 

Hopefully this gives a better insight into my standpoint! I want our core components to be as clear-cut as possible without introducing unnecessary ambiguity.

Elanhant:
Updated the doc and the PR description

PR #80, opened 2019-10-21
fix: use babel runtime transform to avoid issues when importing async/await code

### LINKED ISSUE #79: Issues when using CopyButton component
Trying to use `CopyButton` in an application results in the following error:
>index.es.js:6356 Uncaught ReferenceError: regeneratorRuntime is not defined
>    at index.es.js:6356
>    at CopyButton (index.es.js:6385)

This seems to happen because of the way `async/await` gets transpiled with Babel.

Elanhant:
Fixes #79 by adding `@babel/plugin-transform-runtime` and `@babel/runtime` as suggested in multiple sources: [one](https://stackoverflow.com/a/36821986), [two](https://codingitwrong.com/2018/02/05/await-off-my-shoulders.html), [three](https://risanb.com/posts/regenerator-runtime-is-not-defined/).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,5 +9,8 @@
     "@babel/preset-react",
     "@babel/typescript"
   ],
-  "plugins": ["@babel/plugin-proposal-class-properties"]
+  "plugins": [
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-transform-runtime"
+  ]
 }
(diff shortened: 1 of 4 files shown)


DSchau:
@joshwcomeau _please_ refactor my async/await in the copy button to promises :) Not a significant ergonomic improvement, and certainly not worth the 8Kb!

(just one guy's opinion -- consider it if it's helpful!)

Elanhant:
That's a great point! I'll see how it goes if we use promises instead of async/await. Keeping this PR open for now.

Elanhant:
Closing this in favour of #84

PR #87, opened 2019-10-24
Pricing card update [WIP]

greglobinski:
Needed for https://github.com/gatsby-inc/mansion/pull/5105

### DIFF
--- a/src/components/core/PricingCard/PricingCard.js
+++ b/src/components/core/PricingCard/PricingCard.js
@@ -111,7 +111,7 @@ PricingCard.Frame = ({ children, ...rest }) => {
   return (
     <div
       css={{
-        boxShadow: `0px 2px 4px rgba(46, 41, 51, 0.08), 0px 4px 8px rgba(71, 63, 79, 0.16)`,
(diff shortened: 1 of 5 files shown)


src/components/core/PricingCard/PricingCard.js
```diff
@@ -456,50 +477,30 @@ PricingCard.Cta = ({ children, plan, ...rest }) => {
...
-            borderColor: color ? color : colors.purple[50],
-            ":hover": {
-              color: colors.white,
-              background: color ? color : colors.purple[50],
-              borderColor: color ? color : colors.purple[50],
-            },
-          }}
+          onClick={e => onClick(e, { plan: name })}
+          variant={variant === `SECONDARY` ? `PRIMARY` : `SECONDARY`}
+          css={() => [
```
Elanhant: We don't really need a function here unless I'm missing something 🙂 
We could just use
```jsx
css={[
  { fontWeight: isSelected ? `800` : `normal` },
  // ...
]}
```

Elanhant: Also, do we have tokens for font weights?

greglobinski: @Elanhant We did not use the tokens for `font-weights` on cloud, I updated the `theme` by importing them from `gatsby-desing-tokens`

PR #88, opened 2019-10-24
Add babel/cli dep

joshwcomeau:
I tried running a local build of gatsby-interface and was given an error along the lines of:

> babel: command not found

The default build command, `build`, points to `babel src -d dist`, which would only work if you have it globally installed.

This isn't actually a huge deal since the main way of building it is with `build:rollup`, which doesn't need the CLI utility, but yeah I figure this'll help anyone else trying to build locally =)

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "react-icons": "^3.2.1"
   },
   "devDependencies": {
+    "@babel/cli": "^7.6.4",
     "@babel/core": "^7.0.0",
     "@babel/node": "^7.6.1",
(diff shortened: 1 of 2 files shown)


joshwcomeau:
Hm I wonder, are we actually using the `build` command at all? It's used by the `dev` command, but I feel like we use storybook as our dev environment?

Would love to know if I'm mistaken!

cc @rizchelwood @greglobinski

rizchelwood:
I don't believe we use the `build` command, it's usually just running storybook.

joshwcomeau:
Thanks Rizchel! Closing in favor of #142

PR #91, opened 2019-10-28
TS Portage of modal components

mfrachet:
It's not a finished work but since we made a lot of stuff with @greglobinski, we're looking for some comments and opinions before getting further.

Also looking for some help concerning the #todo section at the end 😄 

The following is a copy of the README file of modals.

---

_This is an imperative way to create modals. Some DOM mutations are involved because of portals._

## API

```jsx
import { ModalCard, ModalProvider } from "gatsby-interface"

const ExamplePanelModal: React.FC<ModalProps> = ({ hideModal }) => (
  <ModalPanel position="right">
    <div> This is a panel modal</div>
    <button onClick={() => hideModal(`card`)}>
      Close another modal than me
    </button>
  </ModalPanel>
)

const ExampleModal: React.FC<ModalProps> = ({ hideTopOfStack }) => (
  <ModalCard>
    <div>This is an example modal</div>
    <button onClick={() => showModal("panel", ExamplePanelModal)}>
      Open a panel modal
    </button>
    <button onClick={hideTopOfStack}>Close me</button>
  </ModalCard>
)

const App = () => (
  <ModalProvider rootId="modal-root">
    <button onClick={() => showModal(`card`, ExampleModal)}>
      Open a Card modal
    </button>
  </ModalProvider>
)
```

## How it works?

At mounting time, the `ModalProvider` will create a DOM root element at the end of the `<body></body>` tag where all of the modals will be rendered. In the previous example, it will create a `<div id="modal-root"></div>` thanks to the `rootId` prop.

Then, the `ModalProvider` will provide methods to all of its consumers to manage the different modals of the application:

```tsx
export interface ModalActions {
  /** Create a modal with a unique name, a Component to render in the modal and some adition options */
  showModal: (
    modalName: string,
    Component: React.FC<ModalProps>,
    options?: ModalOptions
  ) => void
  /** Hide a modal based on its unique name */
  hideModal: (modalName: string) => void
  /** Check if a specific modal is opened */
  isOpened: (modalName: string) => boolean
  /** Close the last element in the stack */
  hideTopOfStack: () => void
}
```

_These functions are passed as props to the `Component` in the `showModal` calls but are also accessible through the `useModal` hook._

By default, the module exposes three modal wrapper or shapes: `ModalCard`, `ModalPanel` and `ModalFullScreen`. You can also create your own one if they don't fit your need.

## TODO

- [ ] Close modal on navigation
- [ ] Animation on closing modals
- [ ] Applying custom animations: What is the preferred way
  - additional classes?
  - Surounding component holding the animation stuff?
  - additional props for custom animations?
  - other?
- [ ] Animate from/to the origin on open/close

## Screenshots

modals


src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  <ModalPanel position={position}>
+    <div>
+      This is a <strong>Panel modal</strong>
+    </div>
+
+    <button onClick={() => hideModal(`panel`)}>Close me</button>
+  </ModalPanel>
+)
+
+const ModalStory = () => {
```
mfrachet: ℹ️ This component is not that important. It's just the storybook story

src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const {
+    showModal,
+    hideModal,
+    isOpened,
+    hideTopOfStack,
+    modals,
+  } = useModalActions(rootRef)
+
+  useLockScroll(Boolean(modals.length))
```
mfrachet: ℹ️ Locks the screen and prevent from scrolling the page when at least one modal is opened

src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useEscapePress(hideTopOfStack)
```
mfrachet: Allows to close the last opened modal using the escape key

src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <ModalContext.Provider
+      value={{ showModal, hideModal, isOpened, hideTopOfStack }}
+    >
+      {children}
+
+      {modals.map(currentModal =>
+        ReactDOM.createPortal(
+          <FocusLock>
```
mfrachet: ℹ️ External dependency component allowing to trap the actual modal concerning the focus mode

src/components/Modal/README.md
```diff
@@ -0,0 +1,72 @@
+_This is an imperative way to create modals. Some DOM mutations are involved because of portals._
```
mfrachet: ℹ️ This is the thing displayed in the content of the PR with markdown compiled to HTML for clarity

src/components/Modal/README.md
```diff
@@ -0,0 +1,72 @@
...
+At mounting time, the `ModalProvider` will create a DOM root element at the end of the `<body></body>` tag where all of the modals will be rendered. In the previous example, it will create a `<div id="modal-root"></div>` thanks to the `rootId` prop.
+
+Then, the `ModalProvider` will provide methods to all of its consumers to manage the different modals of the application:
+
+```tsx
+export interface ModalActions {
+  /** Create a modal with a unique name, a Component to render in the modal and some adition options */
+  showModal: (
+    modalName: string,
+    Component: React.FC<ModalProps>,
```
mfrachet: ℹ️ In the actual application ,it's part of the optional arguments. From what I have in mind, a component is mandatory to be displayed in a modal, so I've forced it a the second argument of the `showModal` function

src/components/Modal/README.md
```diff
@@ -0,0 +1,72 @@
...
+## TODO
+
+- [] Close modal on navigation
+- [] Animation on closing modals
+- [] Applying custom animations: What is the preferred way
+  - additional classes?
+  - Surounding component holding the animation stuff?
+  - additional props for custom animations?
+  - other?
+- [] Animate from/to the origin on open/close
```
mfrachet: I'm interested in your feedback and looking for your help concerning these points 😊

src/components/Modal/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+At mounting time, the `ModalProvider` will create a DOM root element at the end of the `<body></body>` tag where all of the modals will be rendered. In the previous example, it will create a `<div id="modal-root"></div>` thanks to the `rootId` prop.
+
+Then, the `ModalProvider` will provide methods to all of its consumers to manage the different modals of the application:
+
+```tsx
+export interface ModalActions {
+  /** Create a modal with a unique name, a Component to render in the modal and some adition options */
+  showModal: (
+    modalName: string,
```
mfrachet: ℹ️ This is the unique identifier of a modal allowing for any component to close it: `hideModal('uniqueId')`

Elanhant:
```typescript
showModal: (
    modalName: string,
    Component: React.FC<ModalProps>,
    options?: ModalOptions
  ) => void
```
This approach is more suitable for "toast" notifications since those are mostly static and non-interactive. But for modals  it makes it really difficult to update the modal contents from the component that opened it. 

It was quite an inconvenience when I used to work with a similar API, because not only you have to pass some props to the modal (like title, body or onClose handler) but you also have to resubscribe the modal component itself to any dynamic data that it might've just get as its props from the parent component.

I think [this API](https://ui.reach.tech/dialog) provided by Reach UI makes more sense for interactive modal windows.

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { showModal } = useModal()
+  const [type, setType] = useState<ModalOptions["type"]>(`success`)
```
Elanhant: We can use [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) when we want to showcase different values for props 🙂

src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+    clearStack,
+    modals,
+  } = useModalActions(rootRef)
+
+  useLockScroll(Boolean(modals.length))
+  useEscapePress(hideTopOfStack)
+
+  return (
+    <ModalContext.Provider
+      value={{ showModal, hideModal, isOpened, hideTopOfStack }}
```
Elanhant: The `value` should be defined with `React.useMemo`. Otherwise it's going to be changed on every render and cause everything to re-render even if none of its properties have changed.

src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+              clearStack={clearStack}
+              isOpened={isOpened}
+              {...currentModal.opts}
+            />
+          </FocusLock>,
+          currentModal.el
+        )
+      )}
+
+      <div ref={rootRef} />
```
Elanhant: I don't quite understand why do we render this `div` if `useCreateDOMElement` already appends the same element to body?

src/components/Modal/components/ModalCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+    max-width: calc(100% - (${spaces.l} * 2));
+  }
+
+  @media (min-width: ${breakpoints.phablet}px) {
+    width: 620px;
+  }
+`
+
+export interface ModalCardProps extends HTMLProps<HTMLDivElement> {
+  type?: ModalOptions["type"]
```
Elanhant: I feel like this `ModalOptions["type"]` is used quite often in this PR. This is usually a sign that it should be its own type:
```typescript
export type ModalType = "success" | "error" | "info" | "warn"
```
Same goes for `"position"`.

src/components/Modal/components/Overlay.tsx
```diff
@@ -0,0 +1,49 @@
...
+    0% {
+       background-color: transparent
+     }
+   
+     100% {
+       background-color: ${color};
+     }
+   `
+
+const successFade = buildFadeIn(`rgba(55, 182, 53, 0.75)`)
```
Elanhant: Could you please check if the HEX codes for the RGB part of these colors can be found in `palette`? Then we can add these helper functions to transform HEX codes to RGBA and reduce the amount of magic strings:
```typescript
export function hexToRGB(hex) {
  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex)
  return result
    ? {
        r: parseInt(result[1], 16),
        g: parseInt(result[2], 16),
        b: parseInt(result[3], 16),
      }
    : null
}

export function hexToRGBA(hexColor, opacity) {
  const rgbColor = hexToRGB(hexColor)
  if (!rgbColor) {
    return hexColor
  }
  return `rgba(${rgbColor.r}, ${rgbColor.g}, ${rgbColor.b}, ${opacity})`
}
```

src/components/Modal/ModalProvider.tsx
mfrachet: A miss from a refacto 😅 . This has no places here. Removed in the next commit, thanks for the catch 👌

Elanhant:
Have you guys considered [react-spring](https://www.react-spring.io/) for animations? It's an awesome library which allows for more smooth transitions between states.

src/components/Modal/ModalProvider.tsx
mfrachet: Since it's a top-evel component, I've taken this for granted that it won't rerender that much. But if it makes us more confident about re-rendering, it's okay from my side to put a memo on this 😊

src/components/Modal/Modal.stories.tsx
mfrachet: For sure, let me modify the code acorddingly 👌

src/components/Modal/hooks/useCreateDOMElement.ts
```diff
@@ -0,0 +1,13 @@
+import { useRef, useEffect } from "react"
+
+export const useCreateDOMElement = (id: string) => {
+  const domElRef = useRef<HTMLDivElement>(document.createElement(`div`))
```
Elanhant: I think this code might break any SSR since you're referencing `document` without checking if it is available. I'd suggest to use `undefined` as initial value and move the "createElement" part to the effect:
```typescript
const domElRef = useRef<HTMLDivElement>()

useEffect(() => {
  domElRef.current = document.createElement(`div`)
  domElRef.current.id = id
  document.body.appendChild(domElRef.current)

  return () => {
    document.body.removeChild(domElRef.current)
  }
}, [id])

File: src/components/Modal/hooks/useEscapePress.ts [hunk 15]
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from "react"
+
+export const useEscapePress = (callback: () => void) => {
+  useEffect(() => {
+    const keydownHandler = (e: KeyboardEvent) => {
+      if (e.key === `Escape`) {
```
Elanhant: We should start a TypeScript enum somewhere in the code to hold keyboard keys:
```typescript
export enum KeyCode {
  Esc = `Escape`
}
```

src/components/Modal/hooks/useLockScroll.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from "react"
+
+export const useLockScroll = (shouldLock: boolean) => {
+  useEffect(() => {
+    const htmlNode = document.querySelector(`html`)
+    if (htmlNode) {
+      htmlNode.style.overflow = shouldLock ? `hidden` : `auto`
```
Elanhant: What if some other code changed `style.overflow` to a value other than `auto`? We should try to preserve whatever the initial style is and restore it once the lock is off. 

Also, won't this make the page scroll jump to top? I recall having to restore scroll position as well as `overflow` style. I can help out with that if needed 🙂

src/components/Modal/sharedTypes.ts
```diff
@@ -0,0 +1,29 @@
...
+    modalName: string,
+    Component: React.FC<ModalProps>,
+    options?: ModalOptions
+  ) => void
+  /** Hide a modal based on its unique name */
+  hideModal: (modalName: string) => void
+  /** Check if a specific modal is opened */
+  isOpened: (modalName: string) => boolean
+  /** Hide the last element in the stack */
+  hideTopOfStack: () => void
```
Elanhant: `Stack` is an implementation detail; in my opinion we should try to omit those from exposed APIs 🙂 
How about `hideCurrentModal` and `closeAll`?

src/components/Modal/hooks/useCreateDOMElement.ts
mfrachet: Clearing the dom this way is really smart. Didn't thought about that :o

src/components/Modal/hooks/useEscapePress.ts
mfrachet: Do you have any preference for the location of the file?

src/components/Modal/ModalContext.ts
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { ModalOptions, ModalProps } from "./sharedTypes"
+
+export const ModalContext = React.createContext<ModalProps>({
+  showModal: (
+    modalName: string,
+    Component: React.FC<ModalProps>,
+    options?: ModalOptions
+  ) => undefined,
+  hideModal: (modalName: string) => undefined,
```
Elanhant: Generally, if our function does not return anything, it should be typed as `() => void` like in "proper" typed languages where you can have `public void main()`. This makes it more clear that the function should no be expected to return anything (even though in JS it does return `undefined`).

See also [this link](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#callback-types).

mfrachet: Yep, I'm with you on it. But the line you're pointing are not the function interfaces, they are the default values for the ModalContext.

However, I can move them from `undefined` to `{}` (empty body) which may reflect a bit more what you explain here :)

PR #92, opened 2019-10-30
Adding Portal

mfrachet:
A component that allows to render components outside its parent hierarchy.

This goes directly under the discussion people had yesterday concerning creating opinionated accessible components and is necessary to create the declarative Modal component(s).

```jsx
const App = () => (
    <Portal>
        <div>Move me somewhere!</di>
    </Portal>
)
```

---

A note: It seems that Reachui had some difficulties to be displayed on the DevTools because of the two different `document` used by a React app and React Devtools. Since I was able to display my Portals and there content in my Devtools, I didn't manage this behaviour.

However, I think it could be good to keep it there in case somebody gets in trouble so that we have a workaround solution on this: https://github.com/reach/reach-ui/pull/137

### DIFF
--- a/src/components/Portal/README.md
+++ b/src/components/Portal/README.md
@@ -1,0 +1,21 @@
+# Portal
+
+A component that allows to render components outside its parent hierarchy.
+
+## Usage
+
+```jsx
+const App = () => (
+    <Portal>
+        <div>Move me somewhere!</di>
+    </Portal>
+)
+```
+
+By default, this will render the children of the portal in a `gatsby-portal` node at the end of the actual body content. To rely on another name, you can rely on the `tag` prop: `<Portal tag="something-else"><div>Hello world</div></Portal>`.
+
+## Notes
+
+It seems that Reachui had some difficulties to be displayed on the DevTools because of the two different `document` used by a React app and React Devtools. Since I was able to display my Portals and there content in my Devtools, I didn't manage this behaviour.
+
+However, I think it could be good to keep it there in case somebody gets in trouble so that we have a workaround solution on this: https://github.com/reach/reach-ui/pull/137
--- a/src/components/Portal/README.md
+++ b/src/components/Portal/README.md
@@ -16,6 +16,6 @@ By default, this will render the children of the portal in a `gatsby-portal` nod
 
 ## Notes
 
-It seems that Reachui had some difficulties to be displayed on the DevTools because of the two different `document` used by a React app and React Devtools. Since I was able to display my Portals and there content in my Devtools, I didn't manage this behaviour.
+It seems that Reach UI had some difficulties to be displayed on the DevTools because of the two different `document` used by a React app and React Devtools. Since I was able to display my Portals and there content in my Devtools, I didn't manage this behaviour.
 
 However, I think it could be good to keep it there in case somebody gets in trouble so that we have a workaround solution on this: https://github.com/reach/reach-ui/pull/137
--- a/src/components/Portal/index.ts
+++ b/src/components/Portal/index.ts
@@ -5,6 +5,7 @@ export interface PortalProps {
   tag?: string
 }
 
+// Relying on a specific type extending HTMLElement allows to create custom HTML elements like <gatsby-portal></<gatsby-portal>
 interface GatsbyElement extends HTMLElement {}
 
 const Portal: React.FC<PortalProps> = ({ children, tag = `gatsby-portal` }) => {
@@ -19,7 +20,7 @@ const Portal: React.FC<PortalProps> = ({ children, tag = `gatsby-portal` }) => {
 
     return () =>
       portalNodeRef.current && document.body.removeChild(portalNodeRef.current)
-  }, [])
+  }, [tag])
 
   return hasInitialized && portalNodeRef.current
     ? createPortal(children, portalNodeRef.current)
--- a/src/components/Portal/index.ts
+++ b/src/components/Portal/index.ts
@@ -1,0 +1,29 @@
+import React, { useRef, useEffect, useState } from "react"
+import { createPortal } from "react-dom"
+
+export interface PortalProps {
+  tag?: string
+}
+
+interface GatsbyElement extends HTMLElement {}
+
+const Portal: React.FC<PortalProps> = ({ children, tag = `gatsby-portal` }) => {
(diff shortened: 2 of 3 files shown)


src/components/Portal/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { createPortal } from "react-dom"
+
+export interface PortalProps {
+  tag?: string
+}
+
+interface GatsbyElement extends HTMLElement {}
+
+const Portal: React.FC<PortalProps> = ({ children, tag = `gatsby-portal` }) => {
+  const portalNodeRef = useRef<GatsbyElement | undefined>(undefined)
```
mfrachet: ℹ️ Relying on a specific type extending `HTMLElement` allows to create custom HTML elements like `<gatsby-portal></<gatsby-portal>`

src/components/Portal/README.md
```diff
@@ -0,0 +1,21 @@
...
+        <div>Move me somewhere!</di>
+    </Portal>
+)
+```
+
+By default, this will render the children of the portal in a `gatsby-portal` node at the end of the actual body content. To rely on another name, you can rely on the `tag` prop: `<Portal tag="something-else"><div>Hello world</div></Portal>`.
+
+## Notes
+
+It seems that Reachui had some difficulties to be displayed on the DevTools because of the two different `document` used by a React app and React Devtools. Since I was able to display my Portals and there content in my Devtools, I didn't manage this behaviour.
```
joshwcomeau: Nit: My brain reads `Reachui` as a first name, was super confused at first! Maybe `ReachUI` or `Reach UI` would be clearer?

src/components/Portal/index.ts
joshwcomeau: I know I probably sound like a broken record 😅 but I'd love to see this comment added directly as a code comment, maybe just above line 8? Seems like a really useful explanation for why we're extending HTMLElement without adding new properties, and I want developers who miss this PR (or start in the future) to know it :D

src/components/Portal/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  useEffect(() => {
+    portalNodeRef.current = document.createElement(tag)
+    document.body.appendChild(portalNodeRef.current)
+
+    setHasInitialized(true)
+
+    return () =>
+      portalNodeRef.current && document.body.removeChild(portalNodeRef.current)
+  }, [])
```
joshwcomeau: I think we could make this slightly more powerful by making the dependency array `[tag]`; that way, in theory, the parent component could swap out the tag when the props change, and "move" the rendered children from one portal to another.

Granted that's a pretty unlikely edge-case, but I don't think it does any harm =)

mfrachet: Unlikely edge-cases but we often lose users because of edge-cases, thanks for the catch Josh 😊 , I'm putting it in the next commit!

mfrachet:
@joshwcomeau concerning the Toast thing, I think this is a good idea to provide a way to target a unique element. I mean, adding a `to` or `destination` prop wouldn't but that much of work and will probably cover usecases that we will (probably) have in the future if we continue with the a11y library we've discussed yesterday 😊 . I'll tackle it on tomorrow! Thanks for the feedbacks 👌

src/components/Portal/index.ts
```diff
@@ -0,0 +1,63 @@
+import React, { useRef, useState, useLayoutEffect } from "react"
+import { createPortal } from "react-dom"
+
+export interface PortalProps {
+  tag?: string
```
joshwcomeau: How come tag is optional? Won't we always be specifying one?

src/components/Portal/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  target?: string
+}
+
+const useDOMNode = (tag?: string, isUnique?: boolean) => {
+  const domElRef = useRef<HTMLElement | undefined>(undefined)
+
+  useLayoutEffect(() => {
```
joshwcomeau: My understanding is that `useLayoutEffect` fires in the period between layout recalc and paint, in the pixel pipeline, while `useEffect` happens after paint. Assuming my understanding is correct (and it might not be 😅), I'm not sure why we need `useLayoutEffect` here; seems like we could afford to wait a frame before doing this work, and not block rendering for it. But maybe I'm missing something?

I'm also a fan of adding comments whenever I deviate from the "standard" useEffect hook, so that folks working on this component later know why this was needed / if it's still needed after a change.

src/components/Portal/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  return domElRef
+}
+
+const Portal: React.FC<PortalProps> = ({
+  children,
+  tag = `gatsby-portal`,
+  target,
+}) => {
+  const portalNodeRef = useDOMNode(tag)
+  const portalRootRef = useDOMNode(target, true)
```
joshwcomeau: Great idea to abstract this into a hook!

src/components/Portal/index.ts
Elanhant: I think we provide a default value (`gatsby-portal`) judging by readme and `Portal` component below 🙂

src/components/Portal/index.ts
mfrachet: I don't think it's necessary. The reason I tried this way was more about a feeling while playing with the storybook. When jumping between stories, there was like a "laggy" moment between the switch and the actual DOM modifications while relying on `useEffect`.

I don't have that much experience relying on `useLayoutEffect` but I generally use it when playing with DOM nodes manually, so it seems "instant" to the user. Not sure this is the good way to do this. 

And I also think you're right concerning the moment it triggers :)

Some notes from Kent C. Dodds on it: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect :

> Summary
>     - useLayoutEffect: If you need to mutate the DOM and/or DO need to perform measurements
>     - useEffect: If you don't need to interact with the DOM at all or your DOM changes are unobservable (seriously, most of the time you should use this).

I can switch back to `useEffect` but wanted to share this with you 😊 

@joshwcomeau @Elanhant

joshwcomeau: Sorry! Forgot to respond to this, but yeah I'm happy either way :) no change requested. It does make sense to me that if we're mutating the DOM, it _is_ a layout effect.

PR #93, opened 2019-10-30
Adding declarative modals

mfrachet:
Proposal for a primitive concerning `Modal` components

```jsx
<Modal
   type="info"
   aria-label="Some impressive content"
   isVisible={true}
   onClose={handleClose}
 >
    <ModalFullScreen>
      <div>Hello world</div>
    </ModalFullScreen>
</Modal>
```

It relies on https://ui.reach.tech/dialog for now until we decide what we want to do concerning an eventual Gatsby a11y library (see also a first iteration concerning portals https://github.com/gatsby-inc/gatsby-interface/pull/92)

## What's inside?

- `ModalCard` with the same animation as the one in the application
- `ModalPanel` with right / left side animations
- `ModalFullscren` with bottom to top animation
- `Modal` primitive that allows to handle everything concerning a11y (everything can be put as children)
- `type` handling (success, info, warn, error)
- Modal nesting

## The future

- Custom animations (I will make some tries with https://www.react-spring.io/ as suggested by @Elanhant )
- Implementation details could vary depending on the potential a11y library 

## Wondering

- Do you think it could be good to provide a way to lazy render the children (render prop) to avoid undefined props when they depend on a fetching result for example?


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const getBackgroundAnimation = (type?: ModalType) => {
+  if (type === `error`) return errorFade
+  if (type === `warn`) return warnFade
+  if (type === `success`) return successFade
+
+  return infoFade
+}
+
+const Overlay = styled(DialogOverlay)<any>`
```
mfrachet: ℹ️ I'm getting in trouble typing this because of `DialogOverlayProps & { animation: Keyframe }`. Can't get why it doesn't work 😞

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+   `
+
+const successFade = buildFadeIn(hexToRGBA(palette.green[`500`], 0.75))
+const errorFade = buildFadeIn(hexToRGBA(palette.red[`500`], 0.75))
+const infoFade = buildFadeIn(hexToRGBA(palette.purple[`500`], 0.75))
+const warnFade = buildFadeIn(hexToRGBA(palette.orange[`500`], 0.75))
+
+export interface ModalProps {
+  type?: ModalType
+  isVisible?: boolean
```
Elanhant: Can we perhaps use `isOpen` instead of `isVisible`? It would be more in line with Reach UI's Dialog API, and I also feel like "visible" might be incorrectly interpreted as "available in DOM" based on CSS's `visibility: visible` rule.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+       background-color: ${color};
+     }
+   `
+
+const successFade = buildFadeIn(hexToRGBA(palette.green[`500`], 0.75))
+const errorFade = buildFadeIn(hexToRGBA(palette.red[`500`], 0.75))
+const infoFade = buildFadeIn(hexToRGBA(palette.purple[`500`], 0.75))
+const warnFade = buildFadeIn(hexToRGBA(palette.orange[`500`], 0.75))
+
+export interface ModalProps {
```
Elanhant: I think it's better to extend `DialogOverlayProps` from `@reach/dialog` since we are passing props to `DialogOverlay` anyway. That way we get most of the needed types for free while only having to specify `type?: ModalType`.

src/components/Modal/Modal.tsx
Elanhant: OK there are two approaches here:
```typescript
import { Keyframes } from "@emotion/serialize"

const Overlay = styled(DialogOverlay)<{ animation: Keyframes }>`
  animation: ${props => props.animation} 0.5s ease forwards;
`
```
OR
```typescript

const Overlay = styled(DialogOverlay)<{ animation: ReturnType<typeof buildFadeIn> }>`
  animation: ${props => props.animation} 0.5s ease forwards;
`
```
I think that second approach makes more sense in our case because we're not allowing to provide custom animation (for now. at least 😄) and therefore we don't have to avoid coupling animation type with our `buildFadeIn` function.

src/components/Modal/Modal.tsx
mfrachet: Yes, my concerns was more to create our own API and "facade" the actual Reach UI so that we can switch implementation when we'll be ready with a11y 😄 .

Same concern for the isOpen vs isVisible.

Maybe it's over-thought for now 😅

mfrachet:
@Elanhant what do you think of creating another primitive for AltertDialog that proxies Reach UI one while we'll implement our own in the future?

src/components/Modal/Modal.tsx
mfrachet: Thanks for this 👍

Elanhant:
> @Elanhant what do you think of creating another primitive for AltertDialog that proxies Reach UI one while we'll implement our own in the future?

Sure, sounds good!

PR #94, opened 2019-11-04
Create DecorativeDots component

joshwcomeau:
Right now, Cloud uses decorative dots to highlight certain sections on the page / add visual flair. We have 6 individual assets that are used in a number of different ways. I'm working in Cloud to create a "decorative" module that exposes a nicer API for adding these visual elements.

I chatted with @fk about whether each dot asset was specifically tailored for each page, or if we could abstract it out into a smaller subset of reusable bits, and I learned that we aren't married to the exact visuals right now. My initial thought was to create 4 static assets and pick the closest in every scenario, but it occurred to me that this might be a great place to do things generatively!

The nice thing about doing this generatively is that the code remains as simple as having 3-4 static variants, but it's subtly different every time. While most users won't notice, it's a neat little flourish for those that do (and I suspect a significant number of Gatsby users appreciate such details!)

The plan is to still create 3-4 components in Cloud (eg. `StackedCornerDots`, `SubtleSmallDots`, etc) that wrap this lower-level component.

GIF: 

dots-fade


package.json
```diff
@@ -32,8 +32,10 @@
   "dependencies": {
     "@loadable/component": "^5.10.2",
     "@reach/alert": "^0.1.5",
+    "@types/lodash.sample": "^4.2.6",
     "filestack-react": "^3.1.0",
-    "gatsby-design-tokens": "^1.0.0"
+    "gatsby-design-tokens": "^1.0.0",
+    "lodash.sample": "^4.2.1"
```
joshwcomeau: This package adds ~430 bytes gzip to our bundle :D

fk:
> we aren't married to the exact visuals

💯

src/components/DecorativeDots/DecorativeDots.tsx
```diff
@@ -0,0 +1,117 @@
...
+  palette.magenta[200],
+  palette.magenta[500],
+  palette.purple[400],
+  palette.purple[600],
+  palette.teal[200],
+  palette.teal[500],
+  palette.teal[700],
+]
+
+// Every circle is given a random opacity, but it's weighted so that outliers
```
fk: <3

src/components/DecorativeDots/DecorativeDots.helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+    rateOfDecayX,
+    percentageThroughX
+  )
+  const chanceOfBeingBlankY = getChanceOfBeingBlankForAxis(
+    rateOfDecayY,
+    percentageThroughY
+  )
+
+  return mix(chanceOfBeingBlankX, chanceOfBeingBlankY)
+}
```
fk: 🎒📝

Elanhant:
This is brilliant! 👍 
The only concern I have is whether this actually belongs to `gatsby-interface` since it is a bit specific to our app, and if we expect this repo to grow into a proper OSS component library, we'll have to evaluate reusability of all its components 🙂

mfrachet:
Awesome job @joshwcomeau! 👏 

I only have one concern for this PR: moving the utils / presets to their typescript equivalent. I think that other files are also using utils/preset (like the Breadcrumb component). 

I suggest moving them back to JS and maybe rebase when https://github.com/gatsby-inc/gatsby-interface/pull/99 has been merged? 😊

greglobinski:
I really like the idea that we can have a `factory` to produce the Dots. 

I see two issues:
- one is just a feature request/extension. Looking at the newest usage of the Dots, we see that the `fadeStrength`  to turn off/on random dots would not work, I think that we need a mode when a singular dot is although on or off, without any middle fading states.  
 
Screenshot from 2019-11-06 14 44 21@2x

- the second issue relates to the previous one, in cases like above the cloud of dots does a pretty big job to define the visual effect of the page. The cloud looks  like random, but it's not it's drawn by Flo :). I guess that we can improve the algorithm, to make the randomness more predictable, but for now, it's my personal opinion, the ornaments generated by the script are not so nice looking as the ones drawn by Flo. 

just for comparison, in cases like this one, 

Screenshot from 2019-11-06 15 15 43@2x

the importance of the ornament is so low that the 'nice looking'ness is not important factor even for me :)

greglobinski:
@joshwcomeau Would be nice to see how it works on https://staging.gtsb.io/cloud/, maybe all my concerns are unjustified :), if you will not will do a try later

joshwcomeau:
RE: @Elanhant 

> The only concern I have is whether this actually belongs to gatsby-interface since it is a bit specific to our app, and if we expect this repo to grow into a proper OSS component library, we'll have to evaluate reusability of all its components 🙂

Right, yeah. I think for me I'd like to see us move anything that isn't specific to a given Gatsby _product_ (eg. Cloud), but _may_ be specific to Gatsby's brand. I suppose it depends what kind of component library we want to make:

1. One focused on behaviours and agnostic about styles (eg. Reach UI)
2. One with a minimal set of styles meant to be overriden and customized (eg. Bootstrap)
3. One meant as a ready-to-use batteries-included component library with opinionated styles (eg. [Material UI](https://material-ui.com/), [Evergreen](https://evergreen.segment.com/components/search-input))
4. One meant pretty much exclusively for Gatsby products, which we open-source so that folks who are interested can contribute to it or build products that look like Gatsby products, but is not intended to be generic.

Of these, my ideal scenario would be to build both 1. and 4., and have 4 consume 1. I wouldn't really want us to do 2 or 3: there are _so many_ options right now for this, and and it would be so much work to maintain and manage. #1 appeals to me because it's a much more limited scope and the ecosystem really needs it, given how problematic many full-featured libraries are in terms of usability and accessibility. And #4 also seems nice because we get the benefits of an external component library without all the added expectations of the other options.

This PR probably isn't the place for this discussion :) but I'd love to chat about this sometime! Maybe in a couple weeks at our next front-end meeting.

---

RE: @mfrachet 

> I only have one concern for this PR: moving the utils / presets to their typescript equivalent. I think that other files are also using utils/preset (like the Breadcrumb component).

Hm, what is your concern exactly? AFAIK a .js file can still consume a .ts file, since they all get compiled/transpiled to .js anyway. I did a test locally and Breadcrumb still works in my storybook :) 

I appreciate your work in #99, I'll add some comments there :)

---

RE: @greglobinski 

> Looking at the newest usage of the Dots, we see that the fadeStrength to turn off/on random dots would not work, I think that we need a mode when a singular dot is although on or off, without any middle fading states.

Ah I think there may be some confusion - `fadeStrength` controls the gradient of dots, not how faded specific dots are. Here are two examples with contrasting `fadeStrength` values:

Low fadeStrength:

Screenshot 2019-11-06 at 10 22 18 AM

High fadeStrength:

Screenshot 2019-11-06 at 10 22 26 AM

(It's probably not the best name for this prop, admittedly. Suggestions welcome!)

If we absolutely need a variant where all dots are fully opaque, I could add a prop, `randomizeOpacity`, which would toggle between sampling from the `OPACITIES` array, or always using `1`, but part of the goal here is to simultaneously reduce the number of "special" variants while making every mount unique. Which ties in nicely with...

> The cloud looks like random, but it's not, it's drawn by Flo :). I guess that we can improve the algorithm, to make the randomness more predictable, but for now, it's my personal opinion, the ornaments generated by the script are not so nice looking as the ones drawn by Flo.

Yeah, so just for context I should mention that I've been chatting with Flo throughout and he's on-board with all the samples I've shown him! I agree sometimes they don't look as nice - I have ideas for how to improve on that (specifically the biggest issue right now is that it looks too uniform sometimes, since true randomness isn't very aesthetic; I'd like to switch to perlin noise), but I didn't want to invest too much time in this. I expect I'll come back and tweak this when I have some downtime between tasks, because I do think we could make it look nicer.

But yeah, one of the goals for this project was to make this a sort of easter egg; keen-eyed folks will notice that the pattern changes on every refresh, it's exactly the kind of little discovery that would endear me to Gatsby if I discovered it. So I think it's a worthwhile trade-off, even if it doesn't always look as pretty.

> Would be nice to see how it works on https://staging.gtsb.io/cloud/ or https://staging.gtsb.io/pricing/ maybe all my concerns are unjustified :), if you will not I will give it a try later

Hm so I actually think `/cloud` should probably stay as-is :) it has a couple little details that really please me, like the different-sized dots, and this fella:

Screenshot 2019-11-06 at 10 35 07 AM

I think in the future, if I keep hacking on this algorithm / adding little details and customisations, I can imagine us using this everywhere. But in the meantime I'm 100% OK with us deciding that a few pages are worth keeping static, because they have subtle details not easily replicated and/or because they're very prominent pages where we want to be meticulous.

But yeah, you can see the screenshots of the places I've changed in this PR: https://github.com/gatsby-inc/mansion/pull/5278. Flo has approved all of 'em, which I think is a reasonable way to proceed

joshwcomeau:
Gonna merge this but happy to keep discussing / open a new PR if changes are requested!

greglobinski:
@joshwcomeau Thank you for the comprehensive answer :100:

PR #95, opened 2019-11-04
Fix linting

joshwcomeau:
While working on #94, I noticed that somehow I was able to commit a file that hadn't been formatted. I also noticed that there was >100 lint errors when running `eslint`.

In testing, `prettier-eslint` seems to work, but it definitely missed at least 1 thing for me, and apparently it's missed many others. I'm guessing it's flaky.

This change makes the prettier and lint stages exclusive, which I feel like is kind of a good idea anyway. I also fixed the existing lint issues, so that we start with a clean slate :D

### Trying `eslint-plugin-autofix`

I experimented with [eslint-plugin-autofix](https://github.com/aladdin-add/eslint-plugin/tree/master/packages/autofix) to handle cleaning up the ever-so-common "no-unused-vars", but it introduced a bug that actually took me a while to find, so it's clear it's not ready for primetime. For now we'll have to fix them ourselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "dev": "watch 'yarn run build' src",
     "stylelint": "stylelint './src/**/*.js'",
     "format": "prettier --write \"src/**/*.js\"",
-    "lint": "eslint ./src --color --ext .js,.jsx,.ts,.tsx --report-unused-disable-directives",
+    "lint": "eslint ./src --color --ext .js,.jsx,.ts,.tsx",
     "chromatic": "chromatic test",
     "scaffold:icon": "babel-node --presets @babel/env scripts/scaffold-icon/scaffold-icon.js"
   },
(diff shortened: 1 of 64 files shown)


package.json
```diff
@@ -22,8 +22,8 @@
     "build:rollup": "yarn build:types && rollup -c",
     "dev": "watch 'yarn run build' src",
     "stylelint": "stylelint './src/**/*.js'",
-    "format": "prettier-eslint --write \"src/**/*.js\" && eslint src --fix --color --ext .js,.jsx,.ts,.tsx --report-unused-disable-directives",
-    "lint:commit": "prettier-eslint --eslint-config-path=.eslintrc --write",
+    "format": "prettier --write \"src/**/*.js\"",
+    "lint": "eslint ./src --color --ext .js,.jsx,.ts,.tsx",
```
rizchelwood: Should `--fix` be included in the lint script to automatically fix problems?

PR #96, opened 2019-11-04
Create navigation component

amberleyromo:
## Description

Instead of porting existing dotcom navigation code (closing #70), rebuilt the horizontal nav with reusability and accessibility in mind. Of particular note, I spoke with Madalyn and Marcy about more optimal accessible patterns to improve keyboard navigability from the original design. This new component uses a dropdown toggle, which:

- Visually indicates a second level of navigation (right now you have to hover to know which items have subitems)
- Allows toggling the dropdown open/closed via both keyboard and hover.
- Toggle button control is programmatically linked to the subnav.

I also tweaked some of the hover and active styles that were too subtle. Before we ship this to dotcom, will need a design lookover from @fk for:

- active/hover styles
- toggle button design (right now just an upside down caret)

## Notes

- New skeleton component `<BaseNavigation>`.
- New component `<Navigation>`, which inherits from `<BaseNavigation>`.
- Both components have styles split into a separate object import, to de-clutter the main component.
- Can supply nav items as props or children to the `<Navigation>` component.
	- Including buttons (so we don't have to [hack it](https://github.com/gatsby-inc/mansion/blob/195409e9d7ca105fe72ebf6063653bd860a2a3c7/cloud/gatsbyjs.com/src/components/navigation/item.js#L79) 
- Added helpers for `isVisuallyHidden` and `isMobileSafari`
- Changes to presets:
	- Addition to `sizes`
	- Added `transition` file
- Maintained usage of [`activeClassName` approach](https://github.com/gatsby-inc/gatsby-interface/pull/70#discussion_r336175229)
- Decoupled from contentful / source agnostic
- Added stories for:
	- BaseNavigation
		- default
		- hamburger
	- Navigation
		- default (items passed as props)
		- inverted (items passed as props with inverted flag)
		- item links passed as props and children both
		- item links passed as children
		- item links passed as children with nested items
		- item links passed as props with button as child

### Default

default_nav

### Inverted

inverted_nav

### Default with button

w-button_nav

## Questions

_Rebuilding it instead of porting it made most of the [previous questions (and most PR comments)](https://github.com/gatsby-inc/gatsby-interface/pull/70#issue-328393341) irrelevant._

- [ ] Does anyone have strong feelings on whether or not it's actually useful to have this as a component laid over a skeleton component? My feeling is that it works, but it's too complicated to be a skeleton component, and not super worthwhile after all.

## Related issues/PRs

gatsby-inc/mansion#4593
#70


fk:
> - active/hover styles
> - toggle button design (right now just an upside down caret)

Looks great! ✅ 🙏

package.json
```diff
@@ -30,7 +30,8 @@
     "@loadable/component": "^5.10.2",
     "@reach/alert": "^0.1.5",
     "filestack-react": "^3.1.0",
-    "gatsby-design-tokens": "^1.0.0"
+    "gatsby-design-tokens": "^1.0.0",
+    "hex2rgba": "^0.0.1"
```
Elanhant: We have [these helper functions](https://github.com/gatsby-inc/gatsby-interface/blob/dev/src/utils/helpers/hexToRgb.ts) now so we can omit this dependency.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+  Item = BaseNavigation.Item,
+  ItemLink = BaseNavigation.ItemLink,
+  Dropdown = BaseNavigation.Dropdown,
+  DropdownItem = BaseNavigation.DropdownItem,
+  DropdownToggle = BaseNavigation.DropdownToggle,
+  Button = BaseNavigation.Button,
+  ...rest
+}) => {
+  const [isMobileNavOpen, updateMobileNavState] = defaultUpdateMobileNavState
+    ? { defaultIsMobileNavOpen, defaultUpdateMobileNavState }
```
Elanhant: I don't think this code would work, for two reasons:
1. React hooks cannot be applied conditionally ([see React docs](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level))
1. If `defaultUpdateMobileNavState` is present, we're trying to destructure `{ defaultIsMobileNavOpen, defaultUpdateMobileNavState }` as an array, which would result in an error since `[a, b, ...c]` only works for iterables.

I'm not sure on the purpose of `defaultUpdateMobileNavState`, but without it we could just do this:
```javascript
const [isMobileNavOpen, updateMobileNavState] = useState(defaultIsMobileNavOpen)
```
If we do want to provide a way to customize update function, `useReducer` would probably suit better.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+    : useState(false)
+
+  const value = {
```
Elanhant: We should probably check the performance of navigation once we get it in our app; we might have to wrap this context value in `useMemo`.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+  } = BaseNavigation.useNavigationContext()
+
+  return (
+    <button
+      onClick={() => {
+        updateMobileNavState(!isMobileNavOpen)
+      }}
+      aria-expanded={isMobileNavOpen ? `active` : ``}
+      css={{
+        ...baseStyles.hamburger.default,
```
Elanhant: If we aren't expecting any custom `css` here, we can make these a bit cleaner: 
```jsx
css={[
  baseStyles.hamburger.default,
  [mobileNavMediaQuery]: baseStyles.hamburger.mobile,
]}
```
Though this is totally fine too!

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+
+BaseNavigation.Item = ({ item, children, ...rest }) => {
+  const [isDropdownOpen, toggleDropdown] = useState(false)
+  const dropdownItems = item.items || []
+  const dropdownChildren = children || false
+  const itemHasDropdown = dropdownItems.length > 0 || dropdownChildren
+
+  let ref
+  if (itemHasDropdown) {
+    ref = useRef()
```
Elanhant: We cannot use hooks inside conditions, see my comment above 🙂 Same goes for `useOnClickOutside` below

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+    components: { DropdownItem },
+  } = BaseNavigation.useNavigationContext()
+
+  return (
+    <ul
+      css={{
+        ...baseStyles.dropdown(isDropdownOpen),
+      }}
+      // id to associate with aria-controls on BaseNavigation.Item
+      id={`${item.name}-dropdown`}
```
Elanhant: Ideally we should also require an ID for the root navigation component itself. Then we can use it to compose these IDs as well and reduce the chance of multiple elements with the same ID in HTML.

I also like to use library-specific prefixes and BEM notation for IDs generated by our code, something like `$gtsb-nav-${item.name}__dropdown`, but this is a question of preference.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+        // handle closing dropdown on `esc`
+        if (e.keyCode === 27) {
+          toggleDropdown(false)
+        }
+        return
+      }}
+      {...rest}
+    >
+      {dropdownItems.length > 0 &&
+        dropdownItems.map(item => <DropdownItem item={item} />)}
```
Elanhant: Missing `key` prop

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+
+BaseNavigation.DropdownItem = ({ item: { name, linkTo }, ...rest }) => (
+  <li {...rest}>
+    <Link activeClassName="nav-item-active" to={linkTo}>
+      {name}
+    </Link>
+  </li>
+)
+
+BaseNavigation.Button = ({
```
Elanhant: We should probably rename it to `BaseNavigation.LinkButton` so that we don't confuse people using this component.

src/components/skeletons/BaseNavigation/BaseNavigation.js
amberleyromo: It would be easier without it, I agree, but the purpose of `defaultUpdateMobileNavState` is to allow a higher component to pass the update function. For example, on dotcom, that state is shared higher than just the nav area, across the logo/header area. So I wanted the nav component to manage its own state by default, unless it was required to lift it higher.

src/components/skeletons/BaseNavigation/BaseNavigation.js
amberleyromo: Fine with me!

src/components/skeletons/BaseNavigation/BaseNavigation.js
amberleyromo: Agree

src/components/skeletons/BaseNavigation/BaseNavigation.js
Elanhant: We can achieve this like this:
```javascript
// in props destructuring (notice that default values are not provided):
// isMobileNavOpen: customIsMobileNavOpen,
// updateMobileNavState: customUpdateMobileNavState,

const [ownIsMobileNavOpen, ownUpdateMobileNavState] = useState(false)

const isMobileNavOpen = typeof customIsMobileNavOpen !== `undefined`
  ? customIsMobileNavOpen
  : ownIsMobileNavOpen
const updateMobileNavState = typeof customUpdateMobileNavState !== `undefined`
  ? customUpdateMobileNavState
  : ownUpdateMobileNavState
```

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+  Hamburger = BaseNavigation.Hamburger,
+  HamburgerIcon = BaseNavigation.HamburgerIcon,
+  Nav = BaseNavigation.Nav,
+  List = BaseNavigation.List,
+  Item = BaseNavigation.Item,
+  ItemLink = BaseNavigation.ItemLink,
+  Dropdown = BaseNavigation.Dropdown,
+  DropdownItem = BaseNavigation.DropdownItem,
+  DropdownToggle = BaseNavigation.DropdownToggle,
+  Button = BaseNavigation.Button,
```
joshwcomeau: Just thinking out loud... I wonder if this level of customisation is necessary. I think a fair amount of complexity is added because things like `List` and `Nav` can be overridden, which means that we can't make any assumptions about how these components will be used together.

I started typing out a suggestion around using render props, but realized I'd need to spend more time with this problem than I have to sort out if it's a good suggestion. And you've thought about it longer than I have. But yeah, I wonder whether we need this much flexibility, and what opportunities open up if we allow less customisation. For example, we might be able to flatten things a bit, which'd make it more practical to work without using context.

But yeah this is just a stream-of-consciousness, absolutely no change requested :D

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+
+BaseNavigation.HamburgerIcon = ({ ...rest }) => {
+  const { isInverted, isMobileNavOpen } = BaseNavigation.useNavigationContext()
+
+  return (
+    <div
+      className={isMobileNavOpen ? `active` : ``}
+      css={{ ...baseStyles.hamburgerIcon(isInverted) }}
+      {...rest}
+    ></div>
```
joshwcomeau: This is a question for my own education: it doesn't seem like we're doing anything special to mark this up for screen readers. Is the idea that we want this to be invisible to screen readers because they have some other mechanism for opening/closing the menu? Or is this accessible in a way I don't understand?

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+  } = BaseNavigation.useNavigationContext()
+
+  return (
+    <ul
+      css={{
+        ...baseStyles.list.default,
+      }}
+      {...rest}
+    >
+      {items.length > 0 && items.map((item, i) => <Item item={item} />)}
```
joshwcomeau: You'll get a `key` warning when mapping without providing one... and we probably shouldn't use the index since I can imagine items reordering...

I see that items have a `name`, maybe we can use it for the key? I presume names should be unique?

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+          />
+        </Fragment>
+      )}
+    </li>
+  )
+}
+
+BaseNavigation.ItemLink = ({ item, ...rest }) => (
+  <Link activeClassName="nav-item-active" to={item.linkTo} {...rest}>
+    <span>{item.name}</span>
```
joshwcomeau: I'm curious what the `<span>` is for?

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+      onClick={() => {
+        toggleDropdown(!isDropdownOpen)
+      }}
+      css={{
+        ...baseStyles.dropdownToggle(isInverted),
+      }}
+      {...rest}
+    >
+      <span aria-hidden="true">&or;</span>
+      <span css={{ ...visuallyHidden }}>{`${item.name} Menu`}</span>
```
joshwcomeau: I think Aleksei mentioned it earlier, but I don't think we need to spread props that don't need to be shallow-merged. It doesn't really matter (it's not like the performance implications are at all noticeable), but I'm fastidious and I like how tidy `css={visuallyHidden}` is 😅

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+  return (
+    <ul
+      css={{
+        ...baseStyles.dropdown(isDropdownOpen),
+      }}
+      // id to associate with aria-controls on BaseNavigation.Item
+      id={`${item.name}-dropdown`}
+      onKeyDown={e => {
+        // handle closing dropdown on `esc`
+        if (e.keyCode === 27) {
```
joshwcomeau: I super appreciate this comment since who the heck would know the keycode for escape. But, I think we could change this to `ev.code === 'Escape'` for a similar effect (plus, I just learned that [keyCode is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode))

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -0,0 +1,334 @@
...
+        ...baseStyles.button(isInverted),
+      }}
+      {...rest}
+    >
+      {children} {icon && <MdArrowForward />}
+    </LinkButton>
+  )
+}
+
+BaseNavigation.useNavigationContext = () => {
```
joshwcomeau: So one of the "rules of hooks" is that hook names need to start with `use`, so that linters can pick up on them and warn appropriately. And while this example does actually follow that rule, I suspect the linter might not be smart enough to pick up on it, knowing the little bit I know about ASTs.

Plus I've also just never seen a hook used as a method, and I actually missed that it was a hook on my first couple passes.

Might be good to export it separately, `export const useNavigationContext`. A bit more work for importing, but I think it's a worthwhile tradeoff =)

src/components/skeletons/BaseNavigation/BaseNavigationStyles.js
```diff
@@ -0,0 +1,171 @@
...
+}
+
+const HamburgerIconStylesShared = isInverted => {
+  return {
+    borderRadius: 2,
+    height: `3px`,
+    background: isInverted ? colors.white : colors.black,
+    content: `" "`,
+    position: `absolute`,
+    transition: `all 250ms cubic-bezier(.68,-.55,.265,1.55)`,
```
joshwcomeau: Which properties are meant to be transitioned here? From what I can tell it _seems_ to just be `transform` (which is awesome!). I think it'd be good to be explicit about it and set `transition: transform 250ms`

src/components/skeletons/BaseNavigation/BaseNavigationStyles.js
```diff
@@ -0,0 +1,171 @@
...
+      },
+      ":after": {
+        content: JSON.stringify(``),
+        top: 7,
+        width: 16,
+        right: 0,
+        ...HamburgerIconStylesShared(isInverted),
+      },
+      "&.active": {
+        background: `0 0`,
```
joshwcomeau: What does this do? :o

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+}
+
+const DropdownOpenStyles = {
+  fontSize: fontSizes[1],
+  fontFamily: fonts.system.join(`,`),
+  left: `auto`,
+  right: 0,
+  boxShadow: `0px 4px 16px rgba(46, 41, 51, 0.08), 0px 8px 24px rgba(71, 63, 79, 0.16)`,
+  background: colors.white,
+  width: 260,
```
joshwcomeau: I wonder if we'd want this to be adaptive based on the dropdown content?

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+  },
+}
+
+const DropdownMobileStyles = {
+  width: `100%`,
+  color: colors.white,
+  position: `relative`,
+  background: 0,
+  padding: 0,
+  marginBottom: `calc(1.5rem / 2)`,
```
joshwcomeau: Is this done semantically, because it's important to signify that it should be half of 1.5rem, instead of 0.75rem?

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+    return {
+      display: `block`,
+      textDecoration: `none`,
+      color: `inherit`,
+      fontSize: fontSizes[1],
+      fontFamily: fonts.system.join(`,`),
+      fontWeight: `normal`,
+      transition: `opacity ${transition.speed.default}`,
+      WebkitFontSmoothing: `antialiased`,
+      lineHeight: `calc(2 * 5rem / 3)`,
```
joshwcomeau: Similarly I'm not sure why this `calc` exists. Might be good to add a comment =)

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+      position: `relative`,
+      "& span": {
+        position: `relative`,
+      },
```
joshwcomeau: Ah so my concern about broad child selectors like this is that it might be very surprising to someone who supplies a custom child, to see that arbitrary `span`s have a style applied.

I added this as a thing to discuss in our next Cloud Catchup meeting, so I'll save my philosophical musings for them 😅 but maybe in the meantime we can add a data-attribute to the span in question and use that?

Also not a huge priority, I know I'm leaving _a lot_ of comments here and this is not among the most important

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+        content: `" "`,
+        display: `block`,
+        width: 0,
+        height: 1,
+        bottom: -4,
+        opacity: 0.2,
+        background: isInverted
+          ? `linear-gradient(45deg, ${colors.orange[40]}, ${colors.accent})`
+          : `linear-gradient(45deg, ${colors.lilac}, ${colors.gatsby})`,
+        transition: `all ${transition.speed.default}`,
```
joshwcomeau: Similar thought RE: `transition: all`

src/theme/styles/navigation.js
```diff
@@ -0,0 +1,255 @@
...
+      color: colors.grey[50],
+      textDecoration: `none`,
+      padding: `${spaces.s} ${spaces.l}`,
+      display: `block`,
+      transition: `all ${transition.speed.default}`,
+      "&:hover": {
+        color: colors.grey[90],
+        background: hex2rgba(colors.accent, 0.1),
+      },
+      "&:focus-within": {
```
joshwcomeau: TIL `focus-within`! Awesome :o

src/utils/helpers/isMobileSafari.js
```diff
@@ -0,0 +1,15 @@
...
+  if (typeof window === `undefined`) {
+    return false
+  }
+  const { userAgent } = window.navigator
+  const isIOS = /(iPod|iPhone|iPad)/i.test(userAgent)
+  const isWebkit = /AppleWebKit/i.test(userAgent)
+  const isIOSSafari =
+    isIOS && isWebkit && !/(Chrome|CriOS|OPiOS)/.test(userAgent)
+  return isIOSSafari
+}
```
joshwcomeau: This is a cool helper!

But also, it appears to only be used in 1 place, and that place is commented out. Maybe we can remove it?

src/utils/helpers/visuallyHidden.js
```diff
@@ -0,0 +1,12 @@
+// https://a11yproject.com/posts/how-to-hide-content/
+const visuallyHidden = {
+  position: `absolute !important`,
+  height: `1px`,
+  width: `1px`,
+  overflow: `hidden`,
+  clip: `rect(1px 1px 1px 1px)` /* IE6, IE7 */,
+  clip: `rect(1px, 1px, 1px, 1px)`,
+  whiteSpace: `nowrap` /* added line */,
+}
```
joshwcomeau: 👏🏻

We have something like this in Mansion, `src/dashboard/components/shared/VisuallyHidden.js`. Maybe we can export this from `gatsby-interface` and get rid of the one in Mansion?

src/utils/presets/transition.js
```diff
@@ -0,0 +1,13 @@
...
+    fastOutLinearIn: `cubic-bezier(0.4, 0, 1, 1)`,
+  },
+  speed: {
+    snail: `1000ms`,
+    slow: `500ms`,
+    default: `250ms`,
+    fast: `100ms`,
+    blink: `50ms`,
+  },
+}
```
joshwcomeau: yay semantic animations!!

PR #99, opened 2019-11-06
Adding TS declaration file for utils/presets (don't merge)

mfrachet:
I wanted to build the module locally and got in trouble because the `Modal` component relied on the `utils/presets` file which is was not typed.

This PR is a 1-1 declaration file based on the actual utils/preset module and helps building the library (and also rely on utils/presets in TS files)

### DIFF
--- a/src/utils/presets/index.d.ts
+++ b/src/utils/presets/index.d.ts
@@ -1,0 +1,115 @@
+declare module "presets"
+
+/** Colors */
+interface ColorVariant {
+  900: string
+  800: string
+  700: string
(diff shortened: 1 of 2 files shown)


mfrachet:
I wasn't sure about the build time + module resolution between the two languages but with your clarification, I think it's better to keep the TS files and to close this pr 😊

Elanhant:
It's fine to add `*.d.ts` files if it would take too much effort (compared to what you're working on) to rewrite it in TypeScript!

PR #101, opened 2019-11-07
Adding ConfirmationModal

mfrachet:
Adding a ConfirmationModal which is basically a modal card with confirmation / cancelation. Widely inspired by the reach-ui one.

```jsx
 <ConfirmationModal
      initialFocusRef={destructiveRef}
      type="info"
      isOpen={true}
      onDismiss={handleDismiss)}
    >
      <ConfirmationModalLabel>BE CAREFUL!</ConfirmationModalLabel>
      <ConfirmationModalDescription>
        You're about to make something complex!
      </ConfirmationModalDescription>

      <button>Confirmation</button>
      <button ref={destructiveRef}>Destructive</button>
    </ConfirmationModal>
```

---

I'm wondering if `ConfirmationModalLabel` and `ConfirmationModalDescription` shouldn't be part of the `Modal` component and thus can be used elsewhere than only in the `ConfirmationModal` 🤔 which would imply that the ConfirmationModal will only be a Modal with a card inside. Do you have an opinion about it?

### DIFF
--- a/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -1,0 +1,32 @@
+import React, { useRef } from "react"
+import ConfirmationModalContext from "./ConfirmationModalContext"
+import { Modal, ModalCard } from "../Modal"
+import { ModalProps } from "../Modal/Modal"
+
+export interface ConfirmationModalProps extends ModalProps {
+  prefixId?: string
+}
+
+let MODAL_IDS = 0
+export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
+  children,
+  prefixId = `gatsby-modal`,
+  ...props
+}) => {
+  const modalId = useRef<number>(MODAL_IDS++).current
+  const labelId = `${prefixId}-label-${modalId}`
+  const descriptionId = `${prefixId}-description-${modalId}`
+
+  return (
+    <ConfirmationModalContext.Provider value={{ labelId, descriptionId }}>
+      <Modal
+        {...props}
+        role="alertdialog"
+        aria-labelledby={labelId}
+        aria-describedby={descriptionId}
+      >
+        <ModalCard>{children}</ModalCard>
+      </Modal>
+    </ConfirmationModalContext.Provider>
+  )
+}
--- a/src/components/ConfirmationModal/ConfirmationModalComponents.tsx
+++ b/src/components/ConfirmationModal/ConfirmationModalComponents.tsx
@@ -1,0 +1,22 @@
+import React, { useContext } from "react"
+import ConfirmationModalContext from "./ConfirmationModalContext"
+
+export interface ConfirmationModalDetailProps {
+  prefixId?: string
+}
+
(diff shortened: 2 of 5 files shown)


src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useRef } from "react"
+import ConfirmationModalContext from "./ConfirmationModalContext"
+import { Modal, ModalCard } from "../Modal"
+import { ModalProps } from "../Modal/Modal"
+
+export interface ConfirmationModalProps extends ModalProps {
+  prefixId?: string
```
mfrachet: ℹ️ Allows to put something else than `gatsby-modal` as the prefix of the different IDs

src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+let MODAL_IDS = 0
```
mfrachet: ℹ️ Huge side effect that allows to create modals that have a "unique id" (instead of relying on a third party to generate unique ids)

src/components/ConfirmationModal/ConfirmationModalComponents.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return <div id={labelId} {...props} />
+}
+
+export const ConfirmationModalDescription: React.FC<
+  ConfirmationModalDetailProps
+> = ({ prefixId = `gatsby-modal`, ...props }) => {
+  const { descriptionId } = useContext(ConfirmationModalContext)
+
+  return <div id={descriptionId} {...props} />
+}
```
mfrachet: ❓ as mentioned, I'm not sure if it should stay as is or if I should move it to the `Modal` 🤔

src/components/ConfirmationModal/ConfirmationModalContext.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface ConfirmationModalContext {
+  labelId: string
+  descriptionId: string
+}
+
+export default React.createContext<ConfirmationModalContext>({
+  labelId: "",
+  descriptionId: "",
+})
```
mfrachet: ℹ️ These are the (necessary) default values for the context in typescript

src/components/ConfirmationModal/ConfirmationModal.tsx
Elanhant: I'd suggest we just require modal consumers to provide a modal ID in props (similar to how an `id` is required for `htmlFor` to work when it comes to form controls).

Elanhant:
@mfrachet could you please take a look at [this comment](https://github.com/gatsby-inc/gatsby-interface/pull/101#discussion_r343756677)? 

We can't use Symbols here since we have to construct string IDs for a11y, but I think it's reasonable to require component consumers to provide their own IDs.

mfrachet:
From the different switches I've made in mansion, this pr is not relevant for now (at least). If it becomes, will create another one with the new stuff we have.

PR #103, opened 2019-11-07
Form a11y skeletons

Elanhant:
This PR introduces a set of skeleton components with the aim of handling the most crucial a11y relationship for form controls while still supporting custom styling. Supported controls are:
* inputs
* text areas
* selectors
* single checkboxes (like "I agree to Terms&Conditions")
* radio buttons
* checkbox groups

It is also possible to create a custom form field component using `FormField`.


src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,145 @@
...
+  hasHint,
+  children,
+}: FormFieldProps) {
+  const fieldContext = React.useMemo<FormFieldContextValue>(() => {
+    const hintId = getHintId(id)
+    const errorId = getErrorId(id)
+    const controlDescribedBy =
+      [hasError && getErrorId(id), hasHint && getHintId(id)]
+        .filter(describedBy => describedBy)
+        .join(` `) || undefined
```
mfrachet: Not that important: since you already compute `hintId` and `errorId`, you can rely on the already computed value for the comparison:

```jsx
const hintId = getHintId(id)
const errorId = getErrorId(id)

const controlDescribedBy = 
    [hasError && errorId, hasHint && hintId]
        .filter(Boolean)
        .join(` `) || undefined
```

src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,145 @@
...
+  children,
+  validationMode,
+  ...rest
+}) => {
+  const { hasError, meta } = useFormField()
+
+  return (
+    <div
+      id={meta.errorId}
+      aria-live={getErrorAriaLiveAttribute(validationMode)}
```
mfrachet: I didn't know about aria-live! Will use it more often! 👌

joshwcomeau:
Quick meta-question - are we planning on ditching Formik? It seems like this is in many ways a replacement of Formik rather than a wrapper around it.

I definitely wanna go through this in more detail but I figure it'd be good to make sure we're all on the same page around that first =)

cc @rizchelwood @greglobinski @amberleyromo

Elanhant:
> Quick meta-question - are we planning on ditching Formik? It seems like this is in many ways a replacement of Formik rather than a wrapper around it.

Nope, these form components are not intended to be a Formik replacement; they are just compound components that provide basic a11y features for the most common form controls. Therefore they can be used together with Formik (or other form state solutions) and custom styles.
 
> 
> EDIT: So uhh I kinda dislike Formik. I haven't had a really good time using it and it seems like the validation benefits are outweighed by the added complexity. I'm all on-board for doing it our own way!
> 
I'd like to hear more on your experience with Formik! I am personally OK with Formik v1 even though it does have some drawbacks. However, Formik v2 looks much more promising and my plan was to work on a form module for our app built on top of these a11y skeletons using latest Formik.

Also, I am not sure if we really should try to build our own form state solution to replace Formik. From my previous experience with those, it is really difficult to do them right as they tend to grow into unmanageable beasts while not having any real advantages over third-party solutions. I guess I'd feel more confident if we had more dedicated time and resources to work on it, but for now I'd prefer to stick with Formik 🙂

src/components/form/components/FormField.tsx
Elanhant: Oh true!

joshwcomeau:
> I'd like to hear more on your experience with Formik! I am personally OK with Formik v1 even though it does have some drawbacks. However, Formik v2 looks much more promising and my plan was to work on a form module for our app built on top of these a11y skeletons using latest Formik.

So I haven't used v2, maybe it got a lot nicer. I have two big complaints:

- I don't like that it uses context to implicitly pass values. Feels way too magical. The beauty of React is that data flow is explicit; you can always tell why a value is the way it is. When using a `<Field name="username">`, it's totally invisible, you don't know where that value was set. You have to crawl up the tree until you find the parent form.
- The problems it solves aren't problems I have very often. For a registration form or something relatively trivial, I don't find I have any issues using React state, and doing validations myself. It feels like a solution in search of a problem to me. And so because of that it feels like a lot of added complexity for relatively little benefit (the main one seems to be less typing / repetition, but I don't mind those things).

I haven't used it much yet (a little at my previous job, but not a ton), so I'll see how my feelings change over time, and take a look at v2!

Elanhant:
@joshwcomeau I understand your concerns, but I hope that can be mitigated if we don't force the usage of Formik. I, for one, think that it's totally fine to use something simpler like `useState` for forms that don't require any complex logic — as long as we keep these consistent. And my vision for the forms is that we're basically going to have sets of form primitives:

* a11y primitive components from this PR — the most basic components providing the most essential functionality; they can be used to build new controls if necessary
*  styled form controls that @greglobinski is working on — if I understand correctly, they are going to be based on the a11y components
* Formik form controls which I plan to keep in the app codebase for now and which would integrate styled form controls with Formik API to reduce boilerplate

Nothing should stop us from just using the fist two sets (or even just styled form controls) to build forms without Formik.

joshwcomeau:
@Elanhant that sounds great to me! Yeah, if it's opt-in for the usecases that warrant it, I'm all on-board for that.

Elanhant:
Update: I've renamed `*Field` to `*FieldSkeleton` to accommodate for the styled form controls; it makes more sense for those to be named `*Field` as they are going to be used much more often than the a11y skeletons from this PR.

Though ideally we should start considering upgrading to a monorepo, that way we could have different packages like `gatsby-form-skeletons`, `gatsby-form-styled`, `gatsby-form-stateful` etc.

PR #105, opened 2019-11-08
Attempt to fix build + lint on dev

mfrachet:
_Feel free to commit and push on this branch to fix the problems_

---

This morning, I wanted to play with the Modal component on dev in cloud and got in trouble because of the new version of gatsby-interface.

- Modal was not exported in the root index.ts file
- DecorativeDot was not exported in the root index.ts file

I wanted to make a quick PR and build it to play locally and got in trouble while trying to build `dev`. I think that there was a problem with the eslint autofix or something like that.

This PR is an attempt to fix the rollup build + typescript lint.

⚠️ While testing this and using a yarn link, I got in trouble with the `ToastProvider` and a hooks problem like:

Screenshot 2019-11-08 at 16 20 44


.eslintrc
```diff
@@ -87,4 +87,8 @@ overrides:
       "@typescript-eslint/quotes":
         - error
         - backtick
-        - allowTemplateLiterals: true
\ No newline at end of file
+        - allowTemplateLiterals: true
+      "@typescript-eslint/no-unused-vars":
+        - warn
+        - args: "none"
```
mfrachet: ℹ️ This is a quick fix to avoid having `no-unused-variable` when important types for generics like:

```tsx
import { SomeType } from './somewhere';

// SomeType would have thrown an error because it's caught as "unused"
export type Toto = React.FC<SomeType> = () => <div>Hello<div>
```

rollup.config.js
```diff
@@ -36,6 +38,9 @@ export default {
     }),
     resolve({ extensions }),
     commonjs(),
+    postcss({
+      extensions: [`.css`],
+    }),
```
mfrachet: ℹ️ Necessary to build component importing external CSS stuff, like `reach-ui` stylesheet

src/components/skeletons/BaseAnchor/index.ts
```diff
@@ -1,3 +1 @@
-export { default as BaseAnchor } from "./BaseAnchor"
-
-export type BaseAnchorProps = BaseAnchorProps
```
mfrachet: ℹ️ This doesn't make that much sense and I think it's because eslint autofixes 🤔

Elanhant: `export type BaseAnchorProps = BaseAnchorProps` was used so that the type gets exported in the build's `index.ts` file; please make sure that it is true with the switch to `export *` (I think it should work, but it's better be safe 🙂)

src/index.ts
```diff
@@ -96,3 +96,5 @@ export { default as buttonStyles } from "./theme/styles/button"
 export { default as fontWeights } from "./theme/fontWeights"
 
 export * from "./components/icons"
+export * from "./components/DecorativeDots"
```
Elanhant: Please add the same exports to `index-ts-only.ts`

.eslintrc
joshwcomeau: Interesting! Good catch :)

mfrachet:
Please don't merge this 😛 . I'll try to fix the issue with the ToastProvider before 😊

PR #107, opened 2019-11-08
Rewrite icons in TypeScript; add new icons

Elanhant:
I've rewritten `src/icons` to TypeScript and updated the icon scaffold script.

There are also 4 new icons added: `CheckCircleIcon`, `CloseCircleIcon`, `EllipsisIcon` and `InProgressIcon`:
Screen Shot 2019-11-08 at 10 53 51 AM


src/components/icons/IconSkeleton.tsx
```diff
@@ -1,27 +1,21 @@
...
-  medium: `medium`,
-  large: `large`,
-}
-
-const iconHeightBySize = {
-  [ICON_SIZES.xsmall]: `16px`,
-  [ICON_SIZES.small]: `24px`,
-  [ICON_SIZES.medium]: `32px`,
-  [ICON_SIZES.large]: `40px`,
+const iconHeightBySize: Record<IconSize, string> = {
```
joshwcomeau: Ooh I hadn't seen `Record`s before, that's pretty cool! More semantic than `{ [key: IconSize]: string }`.

src/components/icons/icons.stories.tsx
```diff
@@ -72,7 +77,7 @@ const iconBlockCss = css`
 
 const sortedIconComponentNames = Object.keys(icons)
   // filter out __esModule
-  .filter(componentName => typeof icons[componentName] !== `boolean`)
+  .filter(componentName => typeof (icons as any)[componentName] !== `boolean`)
```
joshwcomeau: How come we need `as any`? Could you please update the comment on Line 79 to add some info about this?

src/components/icons/icons.stories.tsx
```diff
@@ -101,7 +106,7 @@ storiesOf(`icons`, module)
 
 sortedIconComponentNames.forEach(componentName => {
   storiesOf(`icons/Single icons`, module).add(componentName, () => {
-    const Component = icons[componentName]
+    const Component = (icons as any)[componentName]
```
joshwcomeau: Ah, I see it's repeated a few times... I wonder if this file needs to be in TS. I've generally not bothered converting tests/stories/anything that doesn't make it to the bundle, since it just seems like more hassle than it's worth. But this is a weakly-held opinion, curious to hear your thoughts!

src/components/icons/icons.ts
```diff
@@ -0,0 +1,9 @@
+export * from "./types"
+export { default as BlogIcon } from "./BlogIcon"
+export { default as CheckIcon } from "./CheckIcon"
+export { default as PortfolioIcon } from "./PortfolioIcon"
+export { default as EcommerceIcon } from "./EcommerceIcon"
+export { default as CheckCircleIcon } from "./CheckCircleIcon"
+export { default as EllipsisIcon } from "./EllipsisIcon"
+export { default as InProgressIcon } from "./InProgressIcon"
+export { default as CloseCircleIcon } from "./CloseCircleIcon"
```
joshwcomeau: I really wish there was a way to generate this dynamically... I've been caught off-guard so many times digging into bugs that were due to an index like this not being updated.

I could do some digging into this idea (exporting all default exports from all valid files in a directory) if you think it has merit, lemme know :)

src/components/icons/types.tsx
```diff
@@ -0,0 +1,9 @@
+export type IconSize = `xxsmall` | `xsmall` | `small` | `medium` | `large`
+
+export type IconSkeletonProps = Omit<JSX.IntrinsicElements['svg'], 'ref'> & {
+  iconName: string;
+  size?: IconSize;
+  applyColorToStroke?: boolean;
+}
```
joshwcomeau: Are we omitting the `ref` because we haven't set up the component to forward it correctly, so we want Typescript to yell at us when we try and pass it one?

It's a pretty busy line, what do you think of declaring an intermediary type, like:

```tsx
type SvgWithoutRef = Omit<JSX.IntrinsicElements['svg'], 'ref'>;

export type IconSkeletonProps = SvgWithoutRef & {
  iconName: string;
  size?: IconSize;
  applyColorToStroke?: boolean;
}
```

(I do realize that for someone with more TS experience, this might feel overly reductive, but I suspect I'm not the only one that finds that line a little hard to parse, I think being able to give that intermediary type a name helps a lot!)

Also, does TS have a spread operator for union types? Flow has something like:

```ts
type IconSkeletonProps = {
  ...SvgWithoutRef;
  iconName: string;
  size?: IconSize;
  applyColorToStroke?: boolean;
}
```

scripts/scaffold-icon/scaffold-icon.js
```diff
@@ -18,7 +18,7 @@ function scaffold(iconName) {
   if (!iconName) {
     console.error(
       `${chalk.red(`Error:`)} No icon name specified, use ${chalk.cyan(
-        `yarn scaffold your-icon-name`
+        `yarn scaffold:icon CheckCircle`
```
joshwcomeau: Good catch!

The fact that this was already broken makes me wonder how often this script is actually run. Are there any docs about what this is / why it's needed? I'd love for us to get into the habit of documenting stuff we come across that takes us a while to understand, so that the next person can develop that understanding more quickly :D

src/components/icons/icons.ts
Elanhant: Well, all these `export` statements are actually added by the script that should be used when adding a new icon (see `yarn scaffold:icon`). I guess we are just missing a warning at the top of the file for the contributors not to touch it.

src/components/icons/types.tsx
Elanhant: > Are we omitting the `ref` because we haven't set up the component to forward it correctly, so we want Typescript to yell at us when we try and pass it one?

It's actually the other way around 🙂 — TypeScript yells at us when spreading `...rest` props on `<svg />` unless we omit `ref`; this is indeed due to our icon components not forwarding refs. So this use of omit is to actually stop it from yelling in the first place and only then to prevent us from passing a ref. 
> 
> It's a pretty busy line, what do you think of declaring an intermediary type, like:
> 

Sure, sounds good 👍 
> 
> Also, does TS have a spread operator for union types? Flow has something like:
> 

Not at the moment, no (though there is [a proposal for that](https://github.com/microsoft/TypeScript/issues/10727))

scripts/scaffold-icon/scaffold-icon.js
Elanhant: I'm a bit lost with this comment, what do you mean by `this was already broken`? 🙂 This single line was more like a typo since it's a console message. The changes to this file only concern upgrading to TS and replacing default script argument notation (`your-icon-name`) to a more clear PascalCase (the old notation still should work).  

As for documentation, what kind is needed here?

PR #109, opened 2019-11-12
Toast components in typescript

mfrachet:
Just converting the Toast component in TypeScript.

_NB: I thought that there was a problem with the ToastProvider, but in fact, it was a problem with multiple versions of React between gatsby-interface (with dev dependencies installed) and .com_


joshwcomeau:
> I thought that there was a problem with the ToastProvider, but in fact, it was a problem with multiple versions of React between gatsby-interface (with dev dependencies installed) and .com

Ahhh, interesting. I just checked and while they're both set to `^16.8.1`, `gatsby-interface` is actually using 16.9 while .com uses 16.8.1. We should remove the `^` to ensure a hard version lock for shared dependencies across projects.

So yeah, we should fix that! Have you tackled that separately, or should I?

src/components/Link/Link.js
```diff
@@ -6,16 +6,8 @@ import { Link as GatsbyLink } from "gatsby"
...
-const Link = ({
-  children,
-  href,
-  target,
-  to,
-  customCss,
-  variant = `DEFAULT`,
-  ...rest
-}) => {
-  showCustomCssDeprecationMessage(customCss)
```
joshwcomeau: hm if we're still passing `customCss` to `commonProps`, should we keep this message?

Let me know if there's context I'm missing around this change =)

src/components/Toast/Toast.tsx
```diff
@@ -81,21 +78,30 @@ const ToastIconByTone = {
...
+export const Toast: React.FC<ToastProps> = ({
   id,
   message,
   tone,
   onRemove,
   closeButtonLabel,
-}) {
-  const IconComponent = ToastIconByTone[tone]
+}) => {
+  const safeTone = tone || `SUCCESS`
```
joshwcomeau: Hm, I wonder if maybe this should be a required prop, instead of setting it to be `SUCCESS`? I checked and it seems in Mansion we always call it with a `tone` (via the hooks).

If we do want to have a default value of `SUCCESS`, I think it'd be good import that value from constants, and to do it via a default param:

```js
export const Toast: React.FC<ToastProps> = ({
  ... stuff
  tone = DEFAULT_TONE,
}) => {
  const IconComponent = ToastIconByTone[tone];

  ... more stuff
}
```

src/components/Toast/hooks.tsx
```diff
@@ -0,0 +1,89 @@
...
+      showToast(
+        <MessageWithLink {...linkProps}>{message}</MessageWithLink>,
+        options
+      )
+    },
+    [showToast]
+  )
+}
+
+export interface IToast {
```
joshwcomeau: "IToast"? What's the "I" for?

src/components/Toast/hooks.tsx
```diff
@@ -0,0 +1,89 @@
...
+  id: number
+  message: string
+  tone: ToastTones
+}
+
+export type ToastInterval = {
+  [key: number]: NodeJS.Timeout
```
joshwcomeau: Ahh this is cool, I hadn't seen `NodeJS.Timeout`!

I googled it and it sounds like maybe this isn't what we want for client-side usage, though: https://stackoverflow.com/a/56970244

mfrachet:
> Ahhh, interesting. I just checked and while they're both set to ^16.8.1, gatsby-interface is actually using 16.9 while .com uses 16.8.1. We should remove the ^ to ensure a hard version lock for shared dependencies across projects.
> 
> So yeah, we should fix that! Have you tackled that separately, or should I?

I ll fix the version here 😊

src/components/Link/Link.js
mfrachet: I don't have much more contexts 😓 . However I think that was to keep the behaviour and providing some feedback without breaking visually. But I didn't see any `customCss` applied on link inside interface nor .com. So it's probably not necessary to keep a reference on `customCss` and thus this function call inside the component.

I'll remove 😊

src/components/Toast/Toast.tsx
mfrachet: I think it could be great to force passing tone for UI oriented component and maybe to created component with a meaningful name inside the host(s?) apps. I'll move 👌 . Thanks for the catch!

src/components/Toast/hooks.tsx
mfrachet: Before moving the hook to a different file, it was conflicting with the `Toast` component 😅 . I'll move back to `Toast`

src/components/Toast/hooks.tsx
mfrachet: I've moved to `number`.  I had to call `setTimeout` on `window` to get a good type matching 😄 .

src/components/Toast/constants.ts
```diff
@@ -0,0 +1,4 @@
+export type ToastTones = "SUCCESS" | "DANGER"
```
Elanhant: Let's rename it to `ToastTone` (singular instead of plural) to keep our enum types names consistent with what we already have (`ButtonSize`, `ButtonVariant` etc). 

Also, unlike a proper enum (i.e. `enum ToastTones`) this type is not a constant and therefore should be exported from a file like `src/components/Toast/types.ts`

src/components/Toast/hooks.tsx
```diff
@@ -0,0 +1,89 @@
...
+  message: string
+  tone: ToastTones
+}
+
+export type ToastInterval = {
+  [key: number]: number
+}
+
+// Side effect to generate a unique toast id and not relying on Math.random
+let TOAST_ID = 0
```
Elanhant: What if we just use `Symbol`? It should work as long as we don't have to manipulate or render it. I'd prefer that we don't introduce any side effects if possible 🙂

src/components/Toast/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./ToastContext"
```
Elanhant: General piece of advice — we should be really careful about wildcard exports in our core library. If we export everything, the library consumers might get access to things that were intended to be internal which leads to at least two problems. Firstly, it adds clutter that makes it more difficult for people to see what they need to import. Secondly, imagine if we remove some function or component that we consider "internal" but which has also been exported via wildcard and used by some application. This presumably small change would result in a broken application! 

So in our case I suggest we only export components and hooks [mentioned in readme](https://github.com/gatsby-inc/gatsby-interface/tree/master/src/components/Toast#components) as well as these types: `ToastTone`, `ToastOptions` (so that consumer apps that use TS can use/extend these when using toasts).

cc: @rizchelwood @greglobinski @joshwcomeau @amberleyromo

Elanhant:
I think it's good to go, great job @mfrachet!

PR #110, opened 2019-11-12
Adding z-index on modal

mfrachet:
Just adding some z-index because the tab title "Organization" in gatsbyjs.com was showing over the modal content 😅


mfrachet:
I merge my fix this way and directly open a PR with a suggestion like yours. It can probably be discussed and challenged, so I simply make it in two times 😊 . Thanks!

PR #111, opened 2019-11-13
Adding Z-indices presets as suggested by @joshwcomeau

mfrachet:
Related comment: https://github.com/gatsby-inc/gatsby-interface/pull/110#pullrequestreview-315645827

Adding z-indices as presets so that we have a clear hierarchy of what can be displayed over what.


src/utils/presets/index.ts
```diff
@@ -3,3 +3,4 @@ export { fontFamilies, fontSizes } from "./typography"
 export { spaces } from "./spaces"
 export { breakpoints } from "./breakpoints"
 export { dimensions, radius } from "./sizes"
+export * from "./zIndices"
```
greglobinski: Can we explicitly say what we exporting. Just for convenience to see what we exporting without need to dig into another file?

mfrachet: Yep 😊  👍

mfrachet:
Thanks for the catch @greglobinski , I  forget the root index.ts export 😬

joshwcomeau:
Awesome :) thanks for this!

> Thanks for the catch @greglobinski , I always forget the root index.ts export 😬

Yeah, I forget about this as well... I wonder if there's anything we can do tooling-wise to help us remember 🤔 I can't think of anything that wouldn't raise a bunch of false-positives.

PR #114, opened 2019-11-14
 Add circleci

rizchelwood:
This adds circleci to `gatsby-interface`

There's a couple of follow up issues I've created to continue improving circleci testing:
- Validate typescript lint rules (I just used the recommended plugin so we should make sure these are the rules we want to enforce) - https://github.com/gatsby-inc/gatsby-interface/issues/115
- Fix jest tests locally then add to circleci (these are failing locally) - https://github.com/gatsby-inc/gatsby-interface/issues/116


src/components/CopyButton/__tests__/CopyButton.test.tsx
```diff
@@ -82,7 +82,7 @@ describe(`<PrimaryButton>`, () => {
 
     fireEvent.click(button)
 
-    let copiedText: string = ``
```
joshwcomeau: Hm, I'm not sure what stripped these types out, but I don't actually think there's anything wrong with them; they're redundant since they can be inferred, but there's no harm in being extra explicit sometimes!

Not sure if this is a setting we can tweak somewhere... ultimately it's not a huge deal though.

.circleci/config.yml
```diff
@@ -0,0 +1,28 @@
...
+          name: Install yarn
+          command: |
+            curl -o- -L https://yarnpkg.com/install.sh
+      - run:
+          name: Install Dependencies
+          command: yarn install
+      - run:
+          name: Build rollup
+          command: yarn build:rollup
+      # TODO: fix tests locally and add them to circleci
```
joshwcomeau: A convention we started at Khan Academy: whenever you add a `# TODO` comment, you had to create an issue and link it (so this comment could become `# TODO (Issue #116): Fix tests locally`).

Just saw you created [an issue](https://github.com/gatsby-inc/gatsby-interface/issues/116) for this, so maybe we could add that number here? Although yeah AFAIK this isn't a convention here so maybe we should discuss as a team first

.eslintrc
```diff
@@ -1,94 +1,35 @@
...
+    ],
+    "eqeqeq": [
+      "warn"
+    ],
+    "no-octal-escape": [
+      "warn"
+    ],
+    "import/no-webpack-loader-syntax": "off",
+    "react-hooks/exhaustive-deps": "off",
+    "react-hooks/rules-of-hooks": "off"
```
joshwcomeau: yay thanks for changing this to json!

I know it wasn't previously enabled, but I'd love to get these two hook rules turned on, they can catch a lot of genuine bugs. Let me know if I should make an issue for this!

src/utils/helpers/copyToClipboard.ts
```diff
@@ -41,3 +23,21 @@ function clipboardWriteTextFallback(content: string): void {
 
   document.body.removeChild(textarea)
 }
+
+export default function copyToClipboard(content: string): Promise<void> {
```
Elanhant: Oh no please don't tell me that we now have a rule that functions must be defined above other functions that use them! 🙏 In my opinion this is pretty harmful as it prevents people from organizing their files in such a way that the main function is always at the top and any helper functions are below it which makes it easier to scan for important information.

.circleci/config.yml
rizchelwood: I think this is a good convention to have and I don't think anyone on the team would oppose so I'll add this 👍

.eslintrc
rizchelwood: I can enable this 👍

src/components/CopyButton/__tests__/CopyButton.test.tsx
rizchelwood: This was autofixed by prettier so I'm not sure exactly what it was. I'll leave this as is for now and we can revisit?

src/utils/helpers/copyToClipboard.ts
rizchelwood: I believe we have this rule enabled in both .com and `gatsby-interface`. That's what I've been used to so I'm okay with the helper functions then main functions format. Maybe we can talk about this as a wider team in our weekly sync to make a decision?

Elanhant: I think questions like lint rules can be discussed asynchronously since they aren't that important to spend sync time on them.

rizchelwood: Agreed!

rizchelwood: From the team discussion, I've disabled this rule

Issue #116, opened 2019-11-14
Fix jest tests and add them to circleci


rizchelwood:
These are currently failing in `dev` (I haven't checked `master`)

PR #118, opened 2019-11-18
chore: fix snapshot testing

Elanhant:
This PR should cover the "fixing" part of #116. It fixes as couple of issues happening during `yarn test`:
* `@reach/portal` used under the hood of our modals was breaking snapshot tests, probably due to  those running on `react-test-renderer` and not having access to refs. Using `render` from `@testing-library/react` as `renderer` option for storyshots addon  helped with this problem.
* Side effect of importing `@reach/dialog/styles.css` wasn't handled by Jest config correctly, I've added this file to `moduleNameMapper` configuration.
* I had to omit stories for Decorative dots from snapshot testing since they are impure and get rendered in a different way each time, which ruins the purpose of snapshots.

I've also updated `react-testing-library` to the latest `@testing-library/react` since there aren't any breaking changes.

### DIFF
--- a/__tests__/Storyshots.test.js
+++ b/__tests__/Storyshots.test.js
@@ -1,5 +1,15 @@
 import initStoryshots from "@storybook/addon-storyshots"
+import { render } from "@testing-library/react"
 
 initStoryshots({
   configPath: `.storybook`,
+  renderer: (storyElement, rendererOptions) => {
+    const { container } = render(storyElement, rendererOptions)
+    return container
+  },
+  /**
+   * We have to exclude DecorativeDots from snapshots since they are always rendered in a differnet way,
+   * which makes snapshot testing useless
+   */
+  storyKindRegex: /^((?!.*?DecorativeDots).)*$/,
 })
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "@storybook/react": "^5.0.11",
     "@storybook/source-loader": "^5.2.0-alpha.14",
     "@svgr/rollup": "^2.4.1",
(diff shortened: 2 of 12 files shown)


__tests__/Storyshots.test.js
```diff
@@ -1,5 +1,15 @@
...
 initStoryshots({
   configPath: `.storybook`,
+  renderer: (storyElement, rendererOptions) => {
+    const { container } = render(storyElement, rendererOptions)
+    return container
+  },
+  /**
+   * We have to exclude DecorativeDots from snapshots since they are always rendered in a differnet way,
+   * which makes snapshot testing useless
+   */
```
joshwcomeau: Ah, sorry about this! I ran into the same issue on Mansion, and my solution there was to avoid rendering this component when `process.env.NODE_ENV` was "test".

Maybe a better solution could be to use predetermined values for the dots when the env is test? That way it still renders the same UI structure, but without the randomness.

Let me know what you think, I can definitely open a separate PR for this myself if you like the idea :+1:

package.json
```diff
@@ -86,6 +86,7 @@
     "@storybook/react": "^5.0.11",
     "@storybook/source-loader": "^5.2.0-alpha.14",
     "@svgr/rollup": "^2.4.1",
+    "@testing-library/react": "^9.3.2",
```
joshwcomeau: We should probably also remove the `react-testing-library` dependency, unless we still need it for some reason?

__tests__/Storyshots.test.js
Elanhant: Sounds good!

PR #120, opened 2019-11-19
chore: introduce commitizen to enforce commit message formatting

Elanhant:
This PR enforces commit messages convention as a first step to enable automated releases. 

It adds [commitizen](https://github.com/commitizen/cz-cli) to the project which allows to easily compose a commit message that follows [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/). By following a commit message convention, we can later automate updates to CHANGELOG based on the commit history.  

#### Example
Screen Shot 2019-11-19 at 9 53 42 AM

Notice that `git commit` is used instead of
```
git commit -m "chore: introduce commitizen to enforce commit message formatting"
```

`commitizen` is run as a Husky hook so it will ask you to compose the commit message even if you use `git commit -m "foo bar"`. It is possible to still use your own message by exiting the process via Ctrl+C, however this is strongly discouraged — ideally, all of our commit messages should follow the convention once it's enabled.

I can add a section on commit messages in README if everyone is OK with the chosen convention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 ## Gatsby Interface
 
+[![Commitizen friendly](https://img.shields.io/badge/commitizen-friendly-brightgreen.svg)](http://commitizen.github.io/cz-cli/)
+
 ![Storybook](https://user-images.githubusercontent.com/18426780/63871208-b9aea600-c978-11e9-9107-79679b699c6f.png)
 
 ```
--- a/package.json
+++ b/package.json
@@ -102,6 +102,8 @@
     "babel-preset-gatsby": "^0.1.11",
     "chalk": "^2.4.2",
     "commander": "^3.0.1",
+    "commitizen": "^4.0.3",
+    "cz-conventional-changelog": "3.0.2",
     "deepmerge": "^4.0.0",
     "dotenv": "^5.0.1",
     "ejs": "^2.7.1",
@@ -155,6 +157,7 @@
   },
   "husky": {
     "hooks": {
+      "prepare-commit-msg": "exec < /dev/tty && git cz --hook || true",
       "pre-commit": "lint-staged"
     }
   },
@@ -168,5 +171,10 @@
       "yarn format",
       "git add"
     ]
+  },
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1140,6 +1140,35 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@commitlint/execute-rule@^8.2.0":
+  version "8.2.0"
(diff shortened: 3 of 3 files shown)


joshwcomeau:
@Elanhant have you used this previously? I'd be keen to hear your experiences. My initial reaction to this idea is honestly pretty negative 😅 but possibly I'm worrying about nothing.

Some concerns:

- This discourages quick "WIP" commits. My workflow is such that I enjoy making quick check-in commits, and I assume they'll be squash-merged when the PR lands. I imagine `--no-verify` will let me skip that, but the onus will be on the developer to fix that later, rather than something that happens automatically while squash-merging.

- I've managed a few projects that follow semver, and my general process for authoring releases involves checking the included changes individually and establishing whether or not they meet the standard of a breaking change, or a minor feature, or entirely invisible to the consumer. I'm a bit worried that having that evaluation happen automatically based on keywords ascribed quickly while working won't be as accurate. When creating a commit, I'm generally not trying to evaluate whether this is a breaking change or not, and I feel like I'll get it wrong more often.

- I perused the docs and it introduces a surprisingly rigorous [specification](https://www.conventionalcommits.org/en/v1.0.0/#specification), which seems problematic given that there's no validation in place. It requires that all developers develop a deep understanding of this spec so that their messages can conform to it, and I'm skeptical that it'll work without some sort of tooling assistance.

- My understanding is that the messages are used to auto-generate changelogs / be used for release notes? I think I'd rather that we author release notes manually. I think the changelog generated could be a useful addendum, but I don't think a straight list of the fixes/features/chores in a set of commits is the most useful summary of a release. It should be editorialized a bit, to focus on the most important changes for someone consuming the library.

My overall impression is that this'll require significant energy from the developers, and it's not clear to me that it's worth the trouble. Authoring releases manually honestly doesn't strike me as a very large source of friction. I do see some upside to having standardized commit messages, but I'm hesitant about any change that can derail a developer in flow, or which discourages frequent committing.

Thinking a little more, points 1 and 2 can be solved by pushing this process to the time when a PR is merged, rather than when a commit is created:

- It still allows WIP commits since they'll all be squashed
- I don't have to think about semver while actively working on features, instead I think about it before merging, when I'm already thinking about the changes at a higher level.

This does have 1 important ramification, which is that PR authors should be the only one to merge them. This is a practice I think we should adopt anyhow, and it conforms with Duffy's [Git Flow](https://www.notion.so/Git-Flow-d93c2afce02a43ad9a5364b7d80fe952) doc.

Curious to hear what other folks think as well. It's entirely possible I'm being fussy. If other folks are onboard, I'm down to give it a shot!

abhiaiyer91:
I think commit formatting and automating the release should be tackled separately. Main goal we need is to get an automated release of this module so we can use them in .com at a faster pace

Elanhant:
Closing this as we're not going to enforce conventional commits for now.

PR #124, opened 2019-11-20
chore: automate prerelease

Elanhant:
This PR introduces automated prereleases for `dev` branch: every commit to `dev` will result in a new "next" version available for installation as `gatsby-interface@next`. 
I added [standard-version](https://github.com/conventional-changelog/standard-version) as a dev dependency to handle the versioning and publishing of the package. It supports the same commit message convention as in #120, but does not enforce it: if there weren't any commits adhering to that convention, it will just bump the patch version. 

I followed [this guide](https://circleci.com/blog/publishing-npm-packages-using-circleci-2-0/) to configure prerelease workflow.

This PR should not be merged until the following is completed (provided that the PR get positive feedback):

- [ ] Replace `chore/automate-prerelease` with `dev` in CircleCI configuration  
- [ ] Remove `--dry-run` flag from `Release "next" version` step in CircleCI configuration


.circleci/config.yml
```diff
@@ -1,30 +1,75 @@
...
+  working_directory: ~/repo
+  docker:
+    - image: circleci/node:10
+
 jobs:
-  build:
-    docker:
-      - image: circleci/node:10
+  is-green:
+    <<: *defaults
```
joshwcomeau: Goodness CircleCI's config is arcane o_O So many symbols.

It might be good to add a comment here explaining this like (I _think_ what it's doing is merging in all default options before this?).

Elanhant:
> * The doc says you need to use an NPM token and set it as an env variable in CircleCI - did you use your personal NPM token, or a gatsby one? It'd be good to add that as a comment too, "Used [dev@gatsbyjs.com](mailto:dev@gatsbyjs.com) NPM token, see credentials in 1password" or something

I might be extremely paranoid, but should we store information like this in an open source library? 😅 Of course, there's nothing *that* sensitive in this particular line, but still...

joshwcomeau:
> I might be extremely paranoid, but should we store information like this in an open source library? 😅 Of course, there's nothing that sensitive in this particular line, but still...

Ah, it's a good point. Maybe we can be a bit more vague: "Used our gatsby NPM account to generate credentials, info stored in our password manager".

I'm not _super_ worried but yeah maybe it's better to be safe than sorry! I'll edit my comment to strip out the name of the password manager. I made that email up so I'm not too worried about it, haha, but on the off-chance it's real (or becomes real later) I'll get rid of it as well.

PR #134, opened 2019-11-22
chore: use 'pure' random values in tests

joshwcomeau:
in #118, @Elanhant had to disable the storyshot generated tests for my DecorativeDots component, since their inherent randomness caused problems.

[As discussed](https://github.com/gatsby-inc/gatsby-interface/pull/118#discussion_r347988080), this fix allows those components to be tested by removing that impurity while `process.env.NODE_ENV` was `test`. This way we can still test their general structure in a repeatable, predictable way.

### DIFF
--- a/__tests__/Storyshots.test.js
+++ b/__tests__/Storyshots.test.js
@@ -7,9 +7,4 @@ initStoryshots({
     const { container } = render(storyElement, rendererOptions)
     return container
   },
-  /**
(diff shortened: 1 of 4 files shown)


Elanhant:
I'm going to switch base to `dev` since our default flow is to merge there first and then merge `dev` to `master` 🙂

joshwcomeau:
> I'm going to switch base to dev since our default flow is to merge there first and then merge dev to master 🙂

Good catch! Thanks

PR #137, opened 2019-11-25
Avatars

### LINKED ISSUE #135: Create an Avatar component

Elanhant:
Fixes #135 and adds two new components: `Avatar` and `AvatarsGroup`.
I also had to fix our TypeScript version to `3.6.2` since using 3.7 breaks our build script:
Screen Shot 2019-11-25 at 10 11 44 AM
 
Note that the size mapping is quite arbitrary at the moment. We'll need to flesh it out together with the design team later.

README copy below:

----
## Avatar and AvatarsGroup

Components to display a single avatar or a group of avatars.

### Avatar

This component can display a single avatar or a fallback if the image is not available

```javascript
import { Avatar } from "gatsby-interface"

function Component() {
  return <Avatar src="https://placekitten.com/200/300" label="A cute kitten" />
}
```

#### Props

| Prop        | Type                        | Default value | Description                                                                                                                     |
| ----------- | --------------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------- |
| src         | `string`                    |               | Required, URL to the avatar image or an empty string if no image available                                                      |
| label       | `string`                    |               | Required, a user-friendly label for the avatar image, usually a person's name                                                   |
| fallback    | `React.ReactNode`           |               | A fallback to be displayed if `src` is an empty string. A common use case for this prob would be to display a person's initials |
| size        | [`AvatarSize`](#avatarsize) | `"medium"`    | An avatar's size                                                                                                                |
| borderColor | `string, null`              | `null`        | When passed, adds a border to the avatar                                                                                        |
| className   | `string`                    |               | Custom class for the avatar element                                                                                             |
| style       | `React.CSSProperties`       |               | Custom style for the avatar element                                                                                             |

##### AvatarSize

```typescript
type AvatarSize = "small" | "medium" | "large" | "xlarge"
```

#### Usage with react-lazy-progressive-image

```jsx
import { Avatar } from "gatsby-interface"
import LazyImage from "react-lazy-progressive-image"

function LazyAvatar({ avatarUrl, userName }) => (
  <LazyImage
    src={avatarUrl || ''}
    // URL for the thumbnail image
    placeholder={avatarUrl && `${avatarUrl}&s=12`}
  >
    {src => (
      <Avatar
        src={src}
        label={userName}
        // Initials for alphabetical writing
        fallback={userName.split(/\s/).map(namePart => namePart.substring(0, 1)).join('')}
      />
    )}
  </LazyImage>
)
```

### AvatarsGroup

This component can be used to display multiple avatars in a single line, e.g. to show a list of contributors or conversation participants.

It also supports an indicator of how many avatars were "truncated"

```javascript
import { AvatarsGroup } from "gatsby-interface"

function Component() {
  return (
    <AvatarsGroup
      avatars={[
        { src: `https://placekitten.com/200/300`, label: `A cute kitten` },
        { src: ``, label: `John Doe`, fallback: "JD" },
        {
          src: `https://loremflickr.com/g/320/240/praha`,
          label: `A random picture of Praha`,
        },
      ]}
      truncatedCount={2}
      truncatedLabel="2 more users"
    />
  )
}
```

**IMPORTANT** Note that `AvatarsGroup` does not handle any slicing of the `avatars` array based on the value passed in `truncatedCount`. That means that if you have a list of 10 avatars but you only want to display 4, you have to slice it yourself:

```javascript
import { AvatarsGroup } from "gatsby-interface"

function Component() {
  const avatars = [
    // 10 avatar descriptors
  ]
  const avatarsToShow = 4
  const truncatedCount = avatars.length - avatarsToShow
  return (
    <AvatarsGroup
      avatars={avatars.slice(0, avatarsToShow)}
      truncatedCount={truncatedCount}
      truncatedLabel={`${truncatedCount} more users`}
    />
  )
}
```

#### Props

| Prop           | Type                                      | Default value              | Description                                                                                                                            |
| -------------- | ----------------------------------------- | -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
| avatars        | [`AvatarDescriptor[]`](#avatardescriptor) |                            | Required, an array of objects describing each avatar in a group                                                                        |
| size           | [`AvatarSize`](#avatarsize)               | `"medium"`                 | What size each avatar in a group should have                                                                                           |
| borderColor    | `string`                                  |                            | Each avatar in a group is bordered; `borderColor` can be used to customize the color of these borders                                  |
| truncatedCount | `number`                                  |                            | When passed an greater than 0, makes the group render an additional "placeholder" avatar that indicates how many users were "left out" |
| truncatedLabel | `string`                                  | `"${truncatedCount} more"` | User-friendly label for the "truncated" placeholder avatar                                                                             |
| className      | `string`                                  |                            | Custom class for the group element                                                                                                     |
| style          | `React.CSSProperties`                     |                            | Custom style for the group element                                                                                                     |

##### AvatarDescriptor

```typescript
export type AvatarDescriptor = Pick<AvatarProps, "src" | "label" | "fallback">
```


src/components/Avatar/AvatarsGroup.tsx
```diff
@@ -0,0 +1,70 @@
...
+        )
+      })}
+      {truncatedCount > 0 && (
+        <Avatar
+          src="" // this is an avatar "lookalike" that indicates how many avatars were truncated
+          label={truncatedLabel}
+          css={overlapCss}
+          {...commonAvatarProps}
+          fallback={`+${truncatedCount}`}
+        />
```
joshwcomeau: Nit: It'd maybe be nice to package this up into its own thin wrapper component:

```tsx
export default function AvatarGroup(...){ ... }

const AvatarTruncatedCount = ({ label, count, size, borderColor }) => {
  const overlapCss = generateOverlapCss(size);

  return (
    <Avatar
      src=""
      css={overlapCss}
      size={size}
      borderColor={borderColor}
      fallback={`+${truncatedCount}`}
    />
  )
}

const generateOverlapCss = size => ({
  marginLeft: `-${borderSizeValues[size] * 2}px`,
});
```

I like this since it moves some of the "implementation details" out of the AvatarGroup body.

But yeah, this one is super subjective, so no change requested =)

src/components/Avatar/README.md
```diff
@@ -0,0 +1,123 @@
...
+  const truncatedCount = avatars.length - avatarsToShow
+  return (
+    <AvatarsGroup
+      avatars={avatars.slice(0, avatarsToShow)}
+      truncatedCount={truncatedCount}
+      truncatedLabel={`${truncatedCount} more users`}
+    />
+  )
+}
+```
```
joshwcomeau: I mentioned earlier that I thought we should include this behaviour. If there's some technical reason I'm not seeing that makes it non-trivial, maybe we could update this comment to explain _why_ it isn't doing this, or possibly create a wrapper component (`TruncatedAvatarsGroup`) that does?

src/components/Avatar/AvatarsGroup.tsx
```diff
@@ -0,0 +1,70 @@
...
+    borderColor,
+  }
+  const overlapCss = {
+    marginLeft: `-${borderSizeValues[size] * 2}px`,
+  }
+  const avatarsShown = avatars.length
+
+  return (
+    <div css={groupBaseCss} className={className} style={style}>
+      {avatars.map(({ src, label, ...avatar }, idx) => {
```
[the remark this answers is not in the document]
Elanhant: Letting the consumer to handle slicing is more flexible than doing it ourselves.

If it's we who does the slicing, the component becomes unusable for cases when the consumer has an already pre-sliced list and a total number. This is similar to pagination where we have a limited set of data and some meta information about what else is available. We'd have to add a new prop (e.g. `truncatedCountToDisplay`) just to handle that specific case. 

Since this is a "core" or "base" component, it's better to leave things like this to the consumers to handle. Kent C. Dodds recently wrote a [nice article on this](https://kentcdodds.com/blog/inversion-of-control), though the approach itself is quite old and I've found it really beneficial while working on my previous projects.

src/components/Avatar/AvatarsGroup.tsx
Elanhant: I feel like this would be an overkill, since most of the added code would be just passing around the same props and defining repetitive types 🙂

src/components/Avatar/README.md
Elanhant: I'll add a comment to the readme based on [my response here](https://github.com/gatsby-inc/gatsby-interface/pull/137#discussion_r350903264). As for adding a wrapper component — this can (and is) done in the app code. I think we should have a really good reason for adding any "wrappers" for the base components in this library to avoid bloating, and for now I'm not entirely convinced that we need a `TruncatedAvatarsGroup`.

src/components/Avatar/AvatarsGroup.tsx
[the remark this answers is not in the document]
joshwcomeau: Ahh, yeah not having the full set of avatars on-hand is a usecase I hadn't considered.

So this gets back to a philosophical thing that still isn't reconciled in my mind - do we want to build components to be generic enough to support usecases we don't expect to need in Cloud? AFAIK we don't plan on paginating avatars, so to design our APIs around a usecase we don't expect to see feels like it could make our lives harder without real benefit... unless we're planning on treating this thing as a first-class open-source citizen, and recommend it as a component library for Gatsby devs (outside of Gatsby Inc)

(Of course, if we _do_ think that we'll be paginating avatars in the future, I withdraw this feedback 😅 though I feel like it'd be good to figure out exactly to what extent we want to value flexibility, and what the expected benefits of that are)

But yeah, so I definitely don't want a third `truncatedCountToDisplay` prop, that feels like the worst of all worlds. I think the fact that the prop is called `truncatedCount` makes me think that it's in charge of actually truncating, and `truncatedLabel` is used exclusively for the displayed text. Maybe we could rename it to `numOfAdditionalAvatars`, `numOfAvatarsNotShown`, or `displayedTruncatedCount`?

Up to you, I don't feel too strongly :+1:

src/components/Avatar/README.md
joshwcomeau: Cool! Yeah, I think all it's missing is the rationale that this is done to support pagination / large numbers of avatars where not all data is available locally

greglobinski:
> type AvatarSize = "small" | "medium" | "large" | "xlarge"

Could we stay consistent with values for `size` prop on any component, following the way we set it for `Button`, S, M, L, XL ?

src/components/Avatar/AvatarsGroup.tsx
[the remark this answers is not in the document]
Elanhant: @joshwcomeau how about `omittedAvatarsCount` and `omittedAvatarsLabel`?

Elanhant:
> > type AvatarSize = "small" | "medium" | "large" | "xlarge"
> 
> Could we stay consistent with values for `size` prop on any component, following the way we set it for `Button`, S, M, L, XL ?

Sure thing! To be honest I prefer the "full" names for such enums, but I'm fine with sticking to what we already have.

PR #141, opened 2019-11-26
Adjusting sidenav style

mfrachet:
This PR aims to create some spacing between the sidenav elements. In fact, when we have elements on more than one line, it's a bit disturbing and it looks like they are two different items.

This PR addresses: https://github.com/gatsby-inc/mansion/issues/5707

The result:

Screenshot 2019-11-26 at 11 24 41

### DIFF
--- a/src/components/SidebarNav/SidebarNav.js
+++ b/src/components/SidebarNav/SidebarNav.js
@@ -3,6 +3,7 @@ import { jsx } from "@emotion/core"
 import { Fragment } from "react"
 import { Link } from "gatsby"
 import PropTypes from "prop-types"
+import { lineHeights } from "gatsby-design-tokens"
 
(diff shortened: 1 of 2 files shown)


src/components/SidebarNav/SidebarNav.js
```diff
@@ -104,7 +104,7 @@ const baseItemStyles = active => {
 
 const baseLinkStyles = {
   color: `inherit`,
-  lineHeight: `1.375rem`,
+  lineHeight: `1.25rem`,
```
fk: https://github.com/gatsbyjs/gatsby/blob/613304cf7f0ab27a69ac9f299c92b492d67665f0/packages/gatsby-design-tokens/src/line-heights.js#L3 has a token that we could use?

PR #143, opened 2019-11-26
Styled InputField [WIP]

greglobinski:


src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </span>
+      )}
+    </FormFieldSkeleton.Label>
+  )
+}
+
+FormField.Label = Label
+FormField.Label.displayName = `FormField.Label`
+
+// todo: find a proper style for
```
greglobinski: @Elanhant Could you please help me set a proper `type` for this

src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+// todo: find a proper type for
+const sharedDescriptionStyles: any = (hasChildren: boolean) => ({
+  alignItems: `flex-start`,
+  color: colors.grey[50],
+  display: `flex`,
+  fontSize: fontSizes[`2xs`],
+  lineHeight: 1.2,
+  margin: `0 ${spaces[`2xs`]}`,
+  marginTop: hasChildren ? spaces.xs : 0,
```
Elanhant: Can we use `marginBottom`? I feel like sticking to one single direction for margins can make our life easier — the label, the control, the hint and the error would all have only `marginBottom` defined.

src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,174 @@
...
+      )}
+    </FormFieldSkeleton.Label>
+  )
+}
+
+FormField.Label = Label
+FormField.Label.displayName = `FormField.Label`
+
+// todo: find a proper type for
+const sharedDescriptionStyles: any = (hasChildren: boolean) => ({
```
Elanhant: The easiest way to define a type here would be this:
```typescript
import { Interpolation } from '@emotion/serialize'

const sharedDescriptionStyles = (hasChildren: boolean): Interpolation => ({
  // ...
})
```

Or you could leverage the `css` function from Emotion to make TS infer the return type (it shouldn't break anything since `css` prop can work with the values returned by that function):
```typescript
import { css } from "@emotion/core"

const sharedDescriptionStyles = (hasChildren: boolean) => css({
  // ...
})
```

src/components/form/components/FormField.tsx
```diff
@@ -0,0 +1,174 @@
...
+  display: `flex`,
+  fontSize: fontSizes[`2xs`],
+  lineHeight: 1.2,
+  margin: `0 ${spaces[`2xs`]}`,
+  marginTop: hasChildren ? spaces.xs : 0,
+  position: `relative`,
+  zIndex: 0,
+
+  svg: {
+    color: `current-color`,
```
Elanhant: I think both `Md*` and our own icons set their color to `currentColor` by default so there's no need for us to specify it :)

src/components/form/components/InputFieldBlock.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <InputField id={id} hasError={!!error} hasHint={!!hint}>
+      <InputField.Wrapper className={className}>
+        <InputField.Label size={labelSize} isRequired={isRequired}>
+          {label}
+        </InputField.Label>
+        <InputField.Control {...rest} />
+        <InputField.Error validationMode={validationMode}>
+          {error}
+        </InputField.Error>
+        <InputField.Hint>{hint}</InputField.Hint>
```
Elanhant: I think that generally hints should be displayed before error messages since hints are more "permanent" information. For example, if we take a password field with a hint like "Your password should contain at least 8 symbols long", it would make sense to display it as close to the input as possible so that the user can easily detect and reference it; and they might get confused if an error appears and pushes the hint down.

src/components/form/stories/form.InputField.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+    options: {
+      showPanel: true,
+    },
+    readme: {
+      sidebar: README,
+    },
+  })
+  .add(`InputField`, () => {
+    const hint = text(`Hint`, ``)
+    const error = text(`Error`, ``)
```
Elanhant: Would you mind adding knobs for `labelSize` here and in `InputFieldBlock` story as well? Take a look at [button size stories](https://github.com/gatsby-inc/gatsby-interface/blob/master/src/components/core/Button/utils/storybook-styles.tsx#L37) for an example.

src/components/form/components/InputField.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { FormField } from "./FormField"
+import InputFieldSkeleton from "../../form-skeletons/components/InputFieldSkeleton"
+import colors from "../../../theme/colors"
+import { fontSizes, radius, spaces } from "../../../utils/presets"
+
+function InputField(props: FormFieldSkeletonProps) {
+  return <InputFieldSkeleton {...props}></InputFieldSkeleton>
+}
+
+const Control: React.FC<
```
Elanhant: Since this is still a wrapper for an `<input />` element, we should forward ref so that consumers can access that element (e.g. to focus on mount):
```typescript
const Control = React.forwardRef<HTMLInputElement, InputFieldSkeletonControlProps>((props, ref) => {
  const { hasError } = FormFieldSkeleton.useFormFieldSkeleton()

  return (
    <InputFieldSkeleton.Control ref={ref} css={{/* ... */}} {...props} />
  )
})
```

src/components/form/components/InputField.tsx
```diff
@@ -0,0 +1,73 @@
...
+        borderRadius: radius.default,
+        color: colors.grey[90],
+        fontSize: fontSizes.s,
+        height: `2.25rem`,
+        padding: `0 ${spaces.s}`,
+        width: `100%`,
+        position: `relative`,
+        zIndex: 1,
+
+        "&:required l": {
```
Elanhant: What is this `l` in the selector? :)

src/components/form/components/FormField.tsx
greglobinski: Good point!

src/components/form/components/InputFieldBlock.tsx
greglobinski: That totally make sense :+1:

src/components/form/components/InputField.tsx
greglobinski: Uppss.

src/components/form/components/FormField.tsx
greglobinski: thank you, I go with `Interpolation`

src/components/form/stories/form.InputField.stories.tsx
greglobinski: Absolutely

src/components/form/components/FormField.tsx
greglobinski: Not 10% the way you suggested, I still will use `marginTop` for `Error`, but true we do not need `hasChildren`. Thanks!

src/components/form/components/InputField.tsx
greglobinski: Right! I also added `forwardRef` to `InputFieldBlock`

greglobinski:
@Elanhant Please help me understand. 

I moved the `enumToOptions` to `src/utils` and when I import it 

```import { enumToOptions } from "../../../utils/helpers"```

in `form.InputField.stories.tsx` it works, but if I import it the same way to `Button/utilis/storybook-styles.tsx` it works on `localhost` but it breaks the CircleCI 

```
yarn run v1.19.1
$ yarn build:types && rollup -c
$ tsc --emitDeclarationOnly -p tsconfig.build-types.json
src/components/core/Button/utils/storybook-styles.tsx:13:31 - error TS7016: Could not find a declaration file for module '../../../../utils/helpers'. '/home/circleci/project/src/utils/helpers/index.js' implicitly has an 'any' type.

13 import { enumToOptions } from "../../../../utils/helpers"
```
https://app.circleci.com/jobs/github/gatsby-inc/gatsby-interface/87

I had to change it to default import to pass the CircleCI
```import enumToOptions from "../../../../utils/helpers/enumToOptions"```

Elanhant:
> @Elanhant Please help me understand.

This happens due to mixing JS and TS — `utils/helpers/index.js` is a JavaScript file so in order to import it to a TypeScript file you either have to rewrite that index file to TS, or add an `utils/helpers/index.d.ts` like I did for [storybook utils](https://github.com/gatsby-inc/gatsby-interface/tree/master/src/utils/storybook).

Using default import from `enumToOptions` works because it is a TypeScript file :)

greglobinski:
> This happens due to mixing JS and TS — utils/helpers/index.js is a JavaScript file so in order to import it to a TypeScript file you either have to rewrite that index file to TS, or add an utils/helpers/index.d.ts like I did for storybook utils .

Thank you Aleksei

greglobinski:
Closed in favor of https://github.com/gatsby-inc/gatsby-interface/pull/158

PR #145, opened 2019-11-27
Fix blank radio issue

joshwcomeau:
Radio buttons have an "emphasized" style, which features a rainbow gradient:

Screenshot 2019-11-27 at 2 18 29 PM

At some point recently, a bug was introduced which caused the contents of the radio button to appear invisible. This is because the radio button uses pseudoelements to create the gradient effect; the `before` element creates a solid gradient backdrop, and the `after` element adds a white backdrop in front of it, cutting in by a few pixels to show the gradient just around the edges, as a border.

The problem is that both of these layers were sitting _in front_ of the primary content, so the white backdrop was blocking the visibility.

This is a quick-fix. There are better solutions (an SVG mask comes to mind, or possibly adding an inner wrapper to the children to not need potentially-problematic negative z-indices), but those can come later.

### DIFF
--- a/__tests__/__snapshots__/Storyshots.test.js.snap
+++ b/__tests__/__snapshots__/Storyshots.test.js.snap
@@ -1026,6 +1026,241 @@ exports[`Storyshots Radio Default 1`] = `
 </div>
 `;
 
+exports[`Storyshots Radio Emphasized 1`] = `
+.emotion-18 {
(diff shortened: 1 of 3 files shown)


joshwcomeau:
I spoke with @Elanhant and he mentioned that I should merge and ship this. Feedback definitely still welcome, can always add a followup PR afterwards!

Issue #139, opened 2019-11-25
Flatten core components folder


Elanhant:
See #90 for more info, but basically we should just move components from "core" subfolder in `src/components` one level up.

mfrachet:
This has been merged on dev, I'm closing for now 😊 . Feel free to reopen!

PR #150, opened 2019-12-02
Flattening core components

mfrachet:
_I'm sorry for the PR size. Some snapshots have been removed / recreated due to a workspace moving (from `core/` to `/`)_

Moving components from `components/core` to `components/`

Since a lot of component have been modified, I will be really thankful if you can play with this branch locally too and verify that it works for you 🙏 

 https://github.com/gatsby-inc/gatsby-interface/issues/139


src/components/Button/Button.stories.tsx
```diff
@@ -67,13 +67,12 @@ storiesOf(`Button`, module)
...
-                ...styles.sizes[`L`],
-                ...styles.variants[`PRIMARY`]({ tone: `BRAND` }),
-              } as any
-            }
+            css={Object.assign(
+              {},
+              styles.base(),
+              styles.sizes.L,
+              styles.variants.PRIMARY({ tone: `BRAND` })
+            )}
```
mfrachet: I had an error like "Spread types may only be created from object types". I've moved to `Object.assign` and it works. Can this be a problem with our TS config?

cc @Elanhant

src/theme/styles/button.d.ts
```diff
@@ -2,11 +2,17 @@
...
 import { ButtonSize, ButtonVariant, ButtonTone } from "../../components/Button"
 
-export const ButtonThemeStyles: {
-  base: (props: { loading?: boolean }) => Interpolation
+const ButtonThemeStyles: {
+  base: (props?: {
+    loading?: boolean
+    leftIcon?: React.ReactNode
+    rightIcon?: React.ReactNode
+  }) => Interpolation
```
mfrachet: I'm not really confident about this piece of code but since we are relying on `leftIcon` and `rightIcon` in the Button component (https://github.com/gatsby-inc/gatsby-interface/blob/6b1e7b75ad63d072018da23a4362d334fe44423a/src/components/Button/Button.tsx#L41) I've added the type definition there. Let me know if it's okay in that way

cc @joshwcomeau  @Elanhant

src/components/AnchorButton/AnchorButton.tsx
```diff
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { jsx } from "@emotion/core"
 import React from "react"
-import { BaseAnchor, BaseAnchorProps } from "../../BaseAnchor"
+import { BaseAnchor, BaseAnchorProps } from "../BaseAnchor"
```
joshwcomeau: I wonder if we should set up aliases in this repo, like we did in gatsby.com... it could have saved a fair amount of trouble with this refactor, looks like!

Maybe something to discuss in our Cloud catch-up tomorrow. Will add to agenda!

src/components/Button/Button.stories.tsx
joshwcomeau: It looks like this is an issue in certain versions of Typescript: https://stackoverflow.com/a/51193091

Not sure what version we're on, but maybe we can look into upgrading?

Ultimately I think it's fine to use `Object.assign` though :) maybe just add a comment to the code so that a developer doesn't try to refactor this to spread props and fall into the same debugging hole.

src/components/Button/Button.tsx
```diff
@@ -36,11 +36,12 @@ export function getButtonStyles({
...
-      ...styles.base({ loading, leftIcon, rightIcon }),
-      ...styles.sizes[size],
-      ...styles.variants[variant]({ tone }),
-    },
+    css: Object.assign(
+      {},
+      styles.base({ loading, leftIcon, rightIcon }),
+      styles.sizes[size],
+      styles.variants[variant]({ tone })
+    ) as any,
```
joshwcomeau: (do we need `as any`? I thought Object.assign was the workaround for TS to work. Not a big deal IMO though)

src/components/Button/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// TODO remove this file once Button is rewritten to TypeScript
+export const Button: any
+export const PrimaryButton: any
+export const SecondaryButton: any
+export const CancelButton: any
+export const SuccessButton: any
+export const TextButton: any
+export const PrimaryDeleteButton: any
+export const SecondaryDeleteButton: any
```
joshwcomeau: Hm, do we need this? It looks like the Button component _is_ already converted to tsx, at least in terms of the file extension. I _think_ we want to keep types there, although TBH I'm not sure. I know this came up in discussion before, but I forget where we landed D:

My personal vote would be to keep all type definitions within `Button.tsx` and `Buttons.tsx` directly. Less confusion, and less potential for bugs (one terrifying thought; what if we add types to `Button.tsx`, but because we have these definitions, the proper types are overwritten with `any`, meaning that the compiler will never complain even when we use the buttons wrong?).

But I'll defer to @Elanhant

src/components/SettingsBlock/SettingsBlock.stories.js
```diff
@@ -87,14 +85,14 @@ storiesOf(`core/SettingsBlock`, module)
             <IntegrationRow
               title="Contentful"
               logoUrl={contentfulLogo}
-              onClickEdit={() => alert(`onClickEdit()`)}
+              onClick={() => alert(`onClickEdit()`)}
```
joshwcomeau: Good catch!! Looks like this prop changed a while ago but the story wasn't updated, appreciate you fixing this :D

One small nit, might be good to change the `alert` as well, `onClick={() => alert('onClick()')}`. Or, for bonus points, you could use `action` instead:

```js
import { action } from "@storybook/addon-actions"

<IntegrationRow
  ...
  onClick={action('onClick')}
/>
```

(totally optional though, you've already done a bunch of good work here so I don't wanna add more to your plate!)

src/theme/styles/button.d.ts
joshwcomeau: Yeah, I'm not sure either... @Elanhant what do you think?

Elanhant: You were perfectly right to add `leftIcon` and `rightIcon` here 👍 

However, there is a seemingly minute, but in fact really important difference in the reasoning for this. With this `button.d.ts` we do not care about the `Button` component — we only care about the `src/theme/styles/button.js` file (in particular, the [base function](https://github.com/gatsby-inc/gatsby-interface/blob/master/src/theme/styles/button.js#L10)) since the `button.d.ts` file is just a declaration file which tells how that `button.js` file can be used with TypeScript. 🙂

src/components/Button/Button.stories.tsx
Elanhant: This is actually an interesting case. The issue is that in `button.d.ts` file we're currently using `Interpolation` as return type for `styles.base` etc while the actual functions return objects. So an obvious fix would be to replace that with an `object`. **However**, I think that it is actually better to leave those types untouched and instead change the styles composition:
```jsx
css={[
  styles.base(),
  styles.sizes.L,
  styles.variants.PRIMARY({ tone: `BRAND` }),
]}
```
This is the Emotion way of composing styles and using it together with the `Interpolation` return type gives us more flexibility.

src/components/Button/index.d.ts
Elanhant: Indeed, I think this file can be safely removed 👍

src/components/AnchorButton/AnchorButton.tsx
mfrachet: This could be totally great and could potentially be linked to https://github.com/gatsby-inc/gatsby-interface/issues/154 . Using this kind of aliases could help us in multiple ways I think

src/components/Button/Button.tsx
mfrachet: looks like the `css` key is waiting for a particular type that I didn't tackle there 😅

src/components/SettingsBlock/SettingsBlock.stories.js
mfrachet: Done 😊

src/theme/styles/button.d.ts
mfrachet: You are right, I was mentioning component in my comment but wanted to reference the component style 😊 . Thanks for the explanation!

__tests__/__snapshots__/Button.js.snap
```diff
@@ -20,7 +20,7 @@ exports[`<PrimaryButton> renders unchanged 1`] = `
   -webkit-box-align: center;
   -ms-flex-align: center;
   align-items: center;
-  border: 1px solid #663399;
+  border: #635e69;
```
Elanhant: This change looks strange 🤔 I guess it came from that `css={[]}` that I suggested to use. Can we verify that the Button still looks the same?

Elanhant: Ah never mind, saw the removed style below, everything should be OK 🙂

PR #156, opened 2019-12-04
Tooltip component

Elanhant:
This PR adds a new component, `Tooltip`. It is based on [Reach UI Tooltip](https://reacttraining.com/reach-ui/tooltip). As always, the component's readme is provided in the end of the PR description.

Some functionality that was changed/added compared to the Reach UI version:
* Wrapping tooltip text
* Tooltip size is limited to window width, i.e. it won't overflow if the tooltip text is really long
The reasoning for these changes is that, while it is a bad UX to have very long tooltips, it is better to safeguard such cases than to punish the end user by having tooltip content cutoff. 

Another change is in the component's API itself. The Reach UI version of Tooltip allows to use it like this:
```jsx
<Tooltip label="Notifications">
  <button style={{ fontSize: 25 }}>
    <span>🔔</span>
  </button>
</Tooltip>
```
It injects all the needed props and refs into `<button />` under the hood. However, I think that this is too "magic" and it is better if we use the render function approach to make it more clear. Although if everyone else is fine with the implicit props&refs injection I can return to the original behaviour.

 
----

Currently only top and bottom positions are supported:
**top**:

tooltip-top

**bottom**:

tooltip-bottom

----

## Tooltip

A component to display a tooltip built on top of [Reach UI Tooltip](https://reacttraining.com/reach-ui/tooltip).

```javascript
import { Tooltip } from "gatsby-interface"

function Component() {
  return (
    <Tooltip label="Lorem ipsum!" position="top">
      {triggerProps => {
        return <button {...triggerProps}>Hover on me!</button>
      }}
    </Tooltip>
  )
}
```

#### Tooltip trigger

A trigger is an element that makes the tooltip appear when it is hovered or focused. It is achieved by spreading the `triggerProps` object in the `Tooltip`'s children render function. It is essential that whatever component you're spreading `triggerProps` on can receive a React ref, and that ref should return a DOM node. This is required to "anchor" the tooltip to its trigger.

#### Tooltips for interactive elements

According to [Reach UI Tooltip](https://reacttraining.com/reach-ui/tooltip#usage), the content inside the trigger is not announced by screenreaders:

> For screenreader users, the only content announced to them is whatever is in the tooltip

If you have some information displayed your trigger that is relevant to the user, you should use `ariaLabel` prop:

```javascript
import { Tooltip } from "gatsby-interface"

function Component() {
  return (
    <Tooltip label="Edit content" ariaLabel="Edit content via Contentful CMS">
      {triggerProps => {
        return <button {...triggerProps}>Contentful CMS</button>
      }}
    </Tooltip>
  )
}
```

#### Touch events

Touch events are not currently supported by [Reach UI Tooltip](https://reacttraining.com/reach-ui/tooltip) and therefor by this component.

### Props

| Prop      | Type                                                | Default value | Description                                                                                                                                       |
| --------- | --------------------------------------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
| label     | `React.ReactNode`                                   |               | Required, the content that is going to be displayed in the tooltip                                                                                |
| ariaLabel | `string`                                            |               | Can be used to override `label` for screenreaders only; a common use case is when the trigger contains some information that you want to announce |
| children  | [`TooltipTriggerRenderFn`](#tooltiptriggerrenderfn) |               | Required, a function that renders the trigger component                                                                                           |
| position  | [`TooltipPosition`](#tooltipposition)               | `"top"`       | Tooltip position relative to the trigger                                                                                                          |

You can also pass any props supported by `<div />` element &mdash; they will be applied to the tooltip.

#### TooltipTriggerRenderFn

```typescript
type TooltipTriggerRenderFn = (
  triggerProps: TooltipTriggerParams
) => React.ReactNode

type TooltipTriggerParams = {
  "aria-describedby": string
  "data-reach-tooltip-trigger": string
  ref: React.Ref<any>
  onMouseEnter: React.ReactEventHandler
  onMouseMove: React.ReactEventHandler
  onFocus: React.ReactEventHandler
  onBlur: React.ReactEventHandler
  onMouseLeave: React.ReactEventHandler
  onKeyDown: React.ReactEventHandler
  onMouseDown: React.ReactEventHandler
}
```

#### TooltipPosition

```typescript
type TooltipPosition = "top" | "bottom"
```


Elanhant:
@fk let me know if you want any changes to how the tooltip looks!

greglobinski:
@Elanhant 

> However, I think that this is too "magic" and it is better if we use the render function approach to make it more clear.

Can we think about use cases when exposing `triggerProps` can help us in any way?

src/components/Tooltip/TooltipPointer.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+function pointerPositionTop(triggerRect: DOMRect) {
+  return {
+    borderLeft: pointerBorderTransparent,
+    borderRight: pointerBorderTransparent,
+    borderTop: pointerBorderFilled,
+    left: getPositionOffsetLeft(triggerRect),
+    top: triggerRect.top - TOOLTIP_POINTER_SIZE + window.scrollY,
+  }
+}
```
joshwcomeau: I noticed when trying this out that on certain screens (maybe non-retina?), there's a flickering, as a subtle sub-pixel gap appears between the tip and the main box:

tip

In the past I've solved this by moving it 1px into the box, so that there's a tiny bit of overlap

mfrachet:
I was saying that I was seeing code that wasn't supposed to still exist in `core/Component`. But I didn't see the conflicts 😓 . I'm waiting for them to be resolved before approving, let me know if you need help to manage these :)

And I find the component really good and it works great on my screens, I don't have that weird pixel mentioned by @joshwcomeau :/

Elanhant:
This PR has too many conflicts and my attempts at merging/rebasing aren't really successful, so to save the time I'm going to close this PR and open a new one (with comments addressed and code updated).

PR #158, opened 2019-12-06
Styled form components

greglobinski:
Screenshot from 2019-12-06 16 14 06@2x

the PR contains https://github.com/gatsby-inc/gatsby-interface/pull/143

### styled fields - styled version of `form-skeletons`:

- `FormField`
- `FormGroupField`
- `InputField`
- `TextAreaField`
- `SelectField`
- `CheckboxField`
- `CheckboxGroupField`
- `RadioButtonField`

### field Blocks - (shortcuts):

- `InputFieldBlock`
- `TextAreaFieldBlock`
- `SelectFieldBlock`
- `CheckboxFieldBlock`
- `CheckboxGroupFieldBlock`
- `RadioButtonFieldBlock`

### connected fields (dedicated to work with `Formik`)

- `InputConnectedField`
- `TextAreaConnectedField`
- `SelectConnectedField`
- `CheckboxConnectedField`
- `CheckboxGroupConnectedField`
- `RadioButtonConnectedField` 

`connnected fields` are the result of my 1to1 with @abhiaiyer91 on Wednesday when he reminded me that we need as simple `api` as we can.  so I prepared another level of abstraction for simple/standard use cases with `Formik`. Although I'm not sure if the components should be maintained in the `gatsby-interface` or in the app.

`connected fields` use `Formik` HoC `connect` to provide formik `state` and `methods` to Field component without manually passing them by `props`

As a side note, maybe it's worth to upgrade Formik on the app to `v2.0` and refactor 'connected' fields to use `useFormikContext` instead of `connect` HoC?

It's still WIP, but I do not want to delay the review any longer. 

ToDo: 
- fix all remaining  TS `type` errors
- ~~solve managing `ref`s for 'group' type Blocks (`CheckboxGroupFieldBlock`, `RadioButtonFieldBlock`)~~
- fix lasting style issues.
- finish documentation


Elanhant:
This looks awesome, @greglobinski! I'll need some time to carefully review, but I'm really happy to see that we're almost there 👍 

> `connnected fields` are the result of my 1to1 with @abhiaiyer91 on Wednesday when he reminded me that we need as simple `api` as we can. so I prepared another level of abstraction for simple/standard use cases with `Formik`. Although I'm not sure if the components should be maintained in the `gatsby-interface` or in the app.

I should've been more clear when discussing forms both with you and Abhi. I was planning to tackle the connected fields myself to free you up for the color presets stuff 🙂 But I'm glad that you've handled it! 

As for the lib vs app concern, I initially planned to keep it in the app. However, we might benefit from having it in gatsby-interface since we'd have type definitions for every component which, if I remember correctly, can be used by IntelliSense even when using these components in vanilla JS.

> 
> As a side note, maybe it's worth to upgrade Formik on the app to `v2.0` and refactor 'connected' fields to use `useFormikContext` instead of `connect` HoC?

This is in my plans! Luckily, there was only one breaking change in Formik v2, and it is really easy to fix.

> * fix all remaining  TS `type` errors

Let me know if you need any help with those. I know how frustrating it can be to setup proper types for form fields and to keep TS compiler happy :)  

> * solve managing `ref`s for 'group' type Blocks (`CheckboxGroupFieldBlock`, `RadioButtonFieldBlock`)

I don't think we really need to support refs for "group" controls. They are not "mirroring" a single HTML element, unlike `InputFIeld` or `SelectField`, and therefore there is no clear definition on what should actually be referenced by the `ref`. 
If we do need a `ref` for things like focus, we could leverage `useImperativeHandle`(https://reactjs.org/docs/hooks-reference.html#useimperativehandle) hook to expose them.

src/components/form/components/CheckboxField.tsx
```diff
@@ -0,0 +1,141 @@
...
+        ref={ref}
+        css={{
+          position: `absolute`,
+          opacity: 0,
+
+          "&:focus + label::before": {
+            ...getFocusStyles(hasError),
+          },
+
+          "&:checked + label::before": {
```
joshwcomeau: These kinds of selectors make me a bit nervous, since they're _super_ hard to reverse-engineer when trying to figure out where a style is coming from. I also have trouble reconciling it with how I think of the component model, where every component owns and controls the DOM it renders, and the same component will look the same no matter where you use it.

That said, I understand why this is used here, because it's a tricky situation; this component renders the input, and whether or not it's focused will affect the sibling state. That behaviour is important / non-negotiable. And the alternatives that I see have significant trade-offs:

- We could mirror focus state in React and pass it down as props or via context, and then just render an icon based on that data. My initial gut reaction to this idea is that it feels gross - copying CSS state to React state always feels bad - but in my experience it actually works really well. It still kinda bothers me, but whenever I've seen this pattern used, it's worked well, and I think it's worth the bad feeling to have the big "traceability" wins, being able to quickly identify why certain CSS is being applied / where it's coming from.
- We could keep both the input and label within the same component. We'd still use the same CSS-only approach but all the HTML elements would be "owned" by the wrapper component:

```js
// Quick doodle, lots of stuff missing
const Wrapper = ({ labelText }) => {
  return (
    <div 
      css={{
        'input:focus + label::before': { styles here },
      }}
    >
      <input />
      <label>{labelText}</label>
    </div>

  )
}
```

This alternative has one big implication, which is that you can no longer independently position the input and labels, they have to be used side-by-side... but I actually kinda think that's a good thing? From a UX / a11y perspective, we probably want a tight coupling between this stuff / not allow developers to accidentally create a bad experience by putting them too far away from each other. But yeah I recognize that this is tricky / maybe there are important usecases this wouldn't work for.

So yeah, because there is no clear winner here IMO, and because this PR is already pretty substantial, I'm not requesting or expecting any changes :) just wanted to bring it up for discussion / see what other folks think.

src/components/form/components/FormField.helpers.tsx
```diff
@@ -0,0 +1,124 @@
...
+)
+
+export const getFocusStyles = (hasError?: boolean) => ({
+  boxShadow: `0 0 0 3px ${hasError ? colors.red[10] : colors.purple[20]}`,
+  borderColor: hasError ? colors.red[30] : colors.purple[40],
+})
+
+export type SharedTextInputStylesOptions = {
+  hasError: boolean | undefined;
+  hasHint: boolean | undefined;
```
joshwcomeau: Nit: I think the slightly-more-conventional way to write this would be:

```ts
hasError?: boolean;
hasHint?: boolean;
```

src/components/form/components/InputConnectedField.tsx
```diff
@@ -0,0 +1,59 @@
...
+          />
+        )
+      }}
+    </Field>
+  )
+}
+
+export default connect(InputConnectedField)
+
+*/
```
joshwcomeau: Sounds like we opted not to go this route? Do we still need this comment?

src/components/form/stories/form.CheckboxField.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+            }
+            labelSize="S"
+            onChange={e => action(`Change`)(e.target.value)}
+            error={error}
+            required={required}
+          />
+        </Wrapper>
+      </StoryUtils.Container>
+    )
+  })
```
joshwcomeau: Great story :+1:

src/components/form/stories/form.CheckboxGroupField.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+            </CheckboxGroupField.Options>
+            <CheckboxGroupField.Hint>{hint}</CheckboxGroupField.Hint>
+            <CheckboxGroupField.Error>{error}</CheckboxGroupField.Error>
+          </CheckboxGroupField>
+        </Wrapper>
+      </StoryUtils.Container>
+    )
+  })
+
+/*
```
joshwcomeau: Another comment I think we can ✂️

src/components/form/components/CheckboxConnectedField.tsx
```diff
@@ -0,0 +1,29 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import { connect, getIn } from "formik"
+import CheckboxFieldBlock from "./CheckboxFieldBlock"
+import Case from "case"
+
+const CheckboxConnectedField = props => {
+  const id = `${props.name}Field`
+  const label = Case.sentence(props.name)
```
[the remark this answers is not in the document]
greglobinski: >  Hadn't heard of this package,
Me too, I found it right now, was looking for something universal.

src/components/form/components/InputConnectedField.tsx
greglobinski: > Do we still need this comment?

no, just my notes :) Thanks for pointing this out.

src/components/form/components/CheckboxField.tsx
greglobinski: Yup, that's a tricky case. 

> We could mirror focus state in React

that would really simplify the logic of shaping the custom input, especially that we already use react `context` on the other hand we will have to manage through context three new values `isHovered`, `isChecked` and `isFocused`. I'm not sure it's worth it. 

> We could keep both the input and label within the same component

`*Field.Control` and `*Field.Label` are static properties of `*Field` component, they are placed in the same file, next to each other, placing them into needless `div` tag only for code readability does not present itself as a solution for me. 

> This alternative has one big implication, which is that you can no longer independently position the input and labels, they have to be used side-by-side

actually you can't independently position them now, the `Label` must follow the `Control` otherwise the `'input:focus + label::before':` selector will not work.

joshwcomeau: > actually, you can't independently position them now, the Label must follow the Control otherwise the 'input:focus + label::before': selector will not work. There is now way to style this way a preceding tag.

Right! That's a good point. It's also a _very_ implicit dependency, one I can imagine biting a developer who doesn't realise. Say if a developer sees this:

```jsx
<Field>
  <Field.Control />
  <Field.Label>I agree</Field.Label>
</Field>
```

They might try and wrap the `Field.Label` in a span, or add an image between the two, or really any sort of tweak to this markup, and they might not realise that they've broken the focus indicators (since they only show up on focus).

IMO that raises the priority of putting them in a wrapper together.

> *Field.Control and *Field.Label are static properties of *Field component, they are placed in the same file, next to each other, placing them into needless div tag only for code readability does not present itself as a solution for me.

Yeah, it's a fair point around discoverability; your first step when looking for the styles on Field.Label would probably be to check the other Field.X components nearby. I think having them share a wrapper would help, since it would be within the same component / same file. But it's not a huge deal.

With this potential for developers accidentally breaking focus styling, I do think it would be worth considering. I worry a bit that it'll break and nobody will notice for months, since it's such a subtle problem. But yeah, maybe it's not as likely as I think.

Elanhant:
I've started looking at this branch locally, going to list my general comments below:

Checkbox description text
----
I feel like it is better if we align the hint/error text with the checkbox itself for checkbox field instead of aligning it with the label: 
Screen Shot 2019-12-09 at 3 40 28 PM
Otherwise the visual "flow" is a bit inconsistent if there are other fields in the form:
Screen Shot 2019-12-09 at 3 42 39 PM

Checkbox group focus styles
----
Seems like focus styles do not work for checkbox groups; on this screenshot the second checkbox is actually focused: 
Screen Shot 2019-12-09 at 3 44 15 PM
 
Checkbox group option wrapping
----
I also think it is better to keep every checkbox group option on its own line as it makes it easier for users to read through them and distinguish one from another

Font sizes
----
This one actually concerns the design team, but I can see at least 3 different font sizes used in the same field here:
Screen Shot 2019-12-09 at 3 57 32 PM
We should probably create a separate issue for this, but I'd like to get @fk opinion on what we can do with the font sizes in our forms? In the PDS that I worked on before we basically had only 2 font sizes (per "size" variant like "medium" or "small") for any form field — one was primary, used for the labels and the input (e.g. 16px) and another one secondary, for the hint/error messages (e.g. 14px)

src/components/form/components/CheckboxField.tsx
greglobinski: Yup, the only way to prevent this is to provide one component `*Field.ControlWithLabel` and withdraw using `*Field.Label` and `*Field.Control` independently 

cc/ @Elanhant

greglobinski:
> I feel like it is better if we align the hint/error text with the checkbox itself for checkbox field instead of aligning it with the label, Otherwise the visual "flow" is a bit inconsistent if there are other fields in the form:

That was the reason why I decided to align the Hint with Label for singular checkboxes, this kicked hint look strange to me, but it's nothing crucial, we can go the consistent way. 
Screenshot from 2019-12-10 09 47 02@2x
 
> Seems like focus styles do not work for checkbox group
:+1: 

> I also think it is better to keep every checkbox group option on its own line as it makes it easier for users to read through them and distinguish one from another

same way like for `Radio` there are two `layout` types for `CheckboxGroup` , vertical and horizontal

Screenshot from 2019-12-10 09 53 25@2x

and we need horizontal layout

Screenshot from 2019-12-10 09 53 58@2x

> I can see at least 3 different font sizes

I set three sizes for `Label`

`M` the default one for `*Field.Label`
Screenshot from 2019-12-10 09 59 08@2x

`L` will be used with `framed` version of `Radio`

Screenshot from 2019-12-10 10 03 22@2x

`S` I added to use for long descriptive labels like in the `Accept Terms` case. We also use this size in EnvVars form

Screenshot from 2019-12-10 10 07 00@2x

but these labels should be increased to `M` size.

Differently than for `*Field.Label` the default size for `*Field.OptionLabel` is set to `L` to unify with text size of other inputs, because in this context they present values.

src/components/form/components/InputConnectedField.tsx
```diff
@@ -0,0 +1,37 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import { connect, getIn } from "formik"
+import InputFieldBlock from "./InputFieldBlock"
+import Case from "case"
+import { InputFieldBlockProps } from "./InputFieldBlock"
+
+export type InputConnectedFieldProps = {
+  name: string;
+  formik?: any; // find better type for this
```
greglobinski: @Elanhant I was not able to find anything better than `any` for `formik`. Tried with `FormikState` and `FormikActions` but for now, have no idea how to use them. Need help

greglobinski:
@fk We would like to here your opinions about the issues pointed out by @Elanhant https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-563496397 but if you see more design issues we should discuss of, maybe as Aleksei suggested, would be better to open a dedicated github issue for design dentered discussion, feel free do choose what better works for you.

Elanhant:
> I set three sizes for Label

Oh, I meant there are three different font sizes displayed in the field:
* label font size (0.875rem)
* the "required" indicator font size (0.75rem)
* an option label font size (1rem)

Sorry, I should've phrased that better 🙂 
There's nothing wrong with having different size "variants" ("M", "L") etc; however, I feel like having three different font sizes in such close proximity makes our field look a bit awkward.

src/components/form/components/InputConnectedField.tsx
Elanhant: Yeah this is a tricky one. We could replace this `any` with `FormikContext` since this is what it is. Problem is, `FormikContext` can be used as a generic type, but I don't think there is a way to preserve this "genericness" when using `connect`.

I would actually suggest to upgrade to Formik v2 in this PR and use its hooks instead of `connect`. I expect that by the time we finish these styled form elements, the app code will also migrate to v2.

src/components/form/components/CheckboxField.tsx
Elanhant: My stance on this is actually a bit different from both options 😄 

I think this is good as it is, and here's my reasoning: it seems sensible to expect that most of the times people will be using the `*FieldBlock` components (and their connected counterparts). Otherwise it would mean that something is wrong with our UI if we have to constantly rely on the granularity of the compound components with their `*Field.Label`, `*Field.Control` etc to adhere to some design spec. 

Therefore we can keep those CSS selectors and just mention in the README that if someone has a dire need to use the compound component, they **have to** render it like this
```jsx
<Field>
  <Field.Control />
  <Field.Label>I agree</Field.Label>
</Field>
```

src/components/form/components/FormField.helpers.tsx
```diff
@@ -0,0 +1,115 @@
...
+  alignItems: `flex-start`,
+  color: colors.grey[50],
+  display: `flex`,
+  fontSize: fontSizes[`2xs`],
+  lineHeight: 1.2,
+  margin: `0 ${spaces[`2xs`]}`,
+  position: `relative`,
+  zIndex: 0,
+
+  svg: {
```
Elanhant: Can we apply these styles directly to `<MdInfoOutline />` and `<MdError />` inside `Hint` and `Error` components? Those components allow to pass node children and therefore we should not target broad selectors like this.

src/components/form/components/InputField.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type InputFieldControlProps = Omit<InputFieldSkeletonControlProps, "ref">
+
+const Control = React.forwardRef<HTMLInputElement, InputFieldControlProps>(
+  (props, ref) => {
+    const { hasError } = FormFieldSkeleton.useFormFieldSkeleton()
+
+    const placeholder =
+      props.placeholder && props.disabled
```
Elanhant: I think a better condition would be `!props.placeholder && props.disabled`, otherwise there's no way to display a custom placeholder (especially if we consider i18n at some point`.

fk:
Thank you @greglobinski that this is still missing input from me. 🙏 :-/

When I started thinking about the indention of the hint text, I took a step back and looked at the indention of `InputField.Label`. Here's a quick shot via devtools showing

1. removing the indention of both label and hint, and removing the default hint icon
2. increasing the current indention
3. the current version

Bildschirmfoto 2019-12-16 um 22 08 57

* What do you think about removing the hint icon? I'm leaning towards that because I think context and visual hierarchy already do that job. Removing makes things easier to scan, at least when removing the indention of the label.
* I changed the label color (of the first two labels) to `grey.90`, thinking that improves visual hierarchy (in general, but especially when both placeholder and hint is present). Can you see anything speaking against making that the new default?

greglobinski:
@fk 

> removing the indention of both label and hint

:+1: that simplifies things

> What do you think about removing the hint icon?

:+1: 

> the label color (of the first two labels) from grey.60 to grey.90

:+1: 

Thank you Flo

fk:
@greglobinski More coming, got sidetracked by a phonecall. Not sure what to do with label sizes and hint text yet, got caught in the middle of that.

fk:
OK, some of the easier things first:

**checkbox/radio group options on their own line**

@greglobinski already answered that one — we have use cases already. Not saying that they are doing the best job, but … they're there. In general I do agree with @Elanhant, and thus was happy to see that `layout="horizontal"` requires opt-in, and each-option-in-its-own-line is the default  (and to have finally checked out a little bit of code). Wouldn't have expected otherwise 💯 👍. 

* On second thought, maybe we should show the default (vertical layout) variant first in the story?
* I also wondered about adding the default value to the Readme. I _think_ it's also not mentioned in the `form-skeletons` Readme, but I might have missed that (or there are other ways to see the (default value for) props in Storybook that I don't know of — noob!).

greglobinski:
@fk

> On second thought, maybe we should show the default (vertical layout) variant first in the story

:+1: 


>  I think it's also not mentioned in the form-skeletons Readme

`form-skeletons` have no styles at all, so there is not `layout` prop on skeletons

fk:
### Checkbox description text

That's actually the difficult thing. Or so I thought. Let's see:
First, I want to make sure I'm not understanding things wrong — we have two different use cases, right?

1. We want to add a hint text to an individual checkbox or radio item to explain that option in more detail. This we want to align with the item label, and we all are in agreement here, correct?
2. We are discussing how to align a hint _or an error_ that applies to the Checkbox/Button _group_, right? An example for an error for a Checkbox group would be "Please check at least two items" … can't really think of an error for a Button group.

Regarding 2), I'm not sure if this is what @Elanhant is describing as the thing where he'd prefer the hint/error text to be aligned with the label of the group options in https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-563496397.

However, I'd say that by removing the hint icon, we should make things quite a bit easier to read, in general. Hints for individual radio/checkbox items can be aligned with their control's label. And removing the indention for hints and Radio/Checkbox group labels, the group hint should cause less visual noise, too — even when appearing along hints for individual group options. (both discussed in https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-566247599 and https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-566273411).

As a follow up, IMO we should

- take a good look whether to allow hints for individual items and hints for groups together.
- think about errors, and hints, that apply to a whole group: Consider a checkbox group that asks to "Please pick at least three items" from a relatively long list of items, let's say 12:
  - Does the user expect this information to show _before_ the list of options? The hint text could be relevant for the action we want them to take, and it could be not even visible, depending on viewport size and number of options.
    - FWIW, I think hint and error works well _below_ `text` and `select` visually/irt scanning. That isn't always the case with radio/checkbox, and even a tall `textarea`.
  - **What other potential A11y issues are we causing by placing the hint text below not only a radio/checkbox group, but a plain _text field_, already?** AFAIU, a screenreader would read the hint after focussing the form field. This can be noisy, or really annoying if the hint reveals that the user needs to go back to correct their input — "You could've told me that _before_ you let me focus the field, and fill it out _wrong._". Looking at the markup, I could see `display: flex` helping here — we could place the hint in between label and input, and move it below the input with `order`: 
image  
    Pinging @marcysutton for help on this — what is the best practise here, are there any resources that can help us?! 🙏 🤗  I wouldn't be surprised we have a lot more to do to really make form validation accessible, but shouldn't hurt starting somewhere, and I think it'd be good to do so now.

fk:
@greglobinski 

> `form-skeletons` have no styles at all, so there is not `layout` prop on skeletons

👍 🤦‍♂ 🙏

greglobinski:
> What other potential A11y issues are we causing by placing the hint text below not only a radio/checkbox group, but a plain text field, already?

AFAIK that is fixed by `aria-describedby` attr on `input` pointing to `hint` and `error` content

fk:
> AFAIK that is fixed by `aria-describedby` attr on `input` pointing to `hint` and `error` content

TIL, and felt shame for being so ignorant about certain things. 🙏

fk:
Last but not least:

### Font sizes

Thanks for calling this out, Aleksei! 🙏 
There's two things I wondered about here:

1 — Currently, the default label and input font-sizes differ — 14px label, 16px input font-size. I liked, and kinda still gravitate towards that slight distinction, I find it helps (me ;-)).
However, I just checked with 16px instead of 14px and _it's not too bad_ — here's a look at that, and all other changes discussed starting with https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-566247599:

Bildschirmfoto 2019-12-17 um 03 21 24

Looking at the `legend` (label) color for radio/checkbox groups, I'm not so sure about the change of the label color any longer. What do you think?

For the sake of completion, here's labels at 14px like we have them now:

Bildschirmfoto 2019-12-17 um 03 24 19

Oh, almost forgot — anticipating 2), I already removed the light purple color for the "required" label. Let's move on to that topic now…

---

2 — I wondered about the current placement of the "required" label — it's placed in the "flow" of the "reading direction", but the eye has to travel quite a bit, and IMO causes a bit of extra visual clutter; how others handle it, and how it might add to the awkward feeling @Elanhant describes in https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-564116936 (and before).

I googled, and (partially re-)read:

* https://www.nngroup.com/articles/required-fields/ — which
  * points out there might be cases which may want us to indicate the _optional_ fields, too
  * resolved my personal fear in using the asterisk prefixed to the label, and my concerns irt using red color for it
* https://dequeuniversity.com/tips/identify-required-elements
* https://www.accessibility-developer-guide.com/examples/forms/required/

All combined, I hacked this together, and I think — while there still are three different font sizes in here — asterisk or not, this works better already IMO:

Bildschirmfoto 2019-12-17 um 03 39 22

As always, curious what you think! 🤗

greglobinski:
Closed in favor of #183

PR #159, opened 2019-12-06
Update navigation component

joshwcomeau:
This PR makes a few small tweaks:

- Aligns nav items to the right instead of left, to match latest designs
- Tweak dropdown positioning
- Adds `Nav.Spacer` to allow items to be pushed to the right
- Code cleanup (mostly carrying over some of the stuff I did in `BaseNavigationStyles`, in a previous PR, to `Navigation`)

Storybook GIF:

story

In-app Screenshots:

Screenshot 2019-12-06 at 4 27 39 PM
Screenshot 2019-12-06 at 4 27 58 PM
Screenshot 2019-12-06 at 4 28 01 PM

### DIFF
--- a/src/theme/styles/navigation.js
+++ b/src/theme/styles/navigation.js
@@ -121,6 +121,7 @@ const DropdownMobileStyles = {
 
 styles.Item = {
   default: {
+    marginBottom: 0,
     padding: `0 ${spaces.s}`,
     "&:hover > ul": {
       ...DropdownOpenStyles,
--- a/src/theme/styles/navigation.js
+++ b/src/theme/styles/navigation.js
@@ -37,8 +37,10 @@ styles.HamburgerIcon = {
 
 styles.Nav = {
   default: {
-    margin: `0 0 0 auto`,
+    display: `flex`,
+    justifyContent: `space-between`,
     padding: `0 ${spaces.m}`,
+    width: "100%",
   },
   mobile: isMobileNavOpen => {
     return {
@@ -61,7 +63,10 @@ styles.Nav = {
 
 styles.List = {
   default: {
+    display: `flex`,
     listStyle: `none`,
+    width: `100%`,
+    alignItems: "center",
   },
   mobile: {
     listStyle: `none`,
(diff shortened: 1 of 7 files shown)


src/theme/styles/navigation.js
```diff
@@ -79,18 +84,16 @@ styles.List = {
 const DropdownOpenStyles = {
   fontSize: fontSizes[1],
   fontFamily: fonts.system.join(`,`),
-  left: `auto`,
   right: 0,
   boxShadow: `0px 4px 16px rgba(46, 41, 51, 0.08), 0px 8px 24px rgba(71, 63, 79, 0.16)`,
   background: colors.white,
-  width: 260,
+  width: 450,
```
Elanhant: Is there a particular reason why we can't use `maxWidth` here? I don't feel really confident about relying on fixed widths in most cases, unless it's something that has to be consistent like a sidebar.

greglobinski: I agree the `width` of dropdown should be adaptable to the content, we can add `max-width` or `min-width` (for really short links), but generally the block should gets its width from the width of content

greglobinski: and then with  custom-fit blocs would not be nice to have the blocks and the beaks centered to the menu link item? instead of this fixed left placed one?  

Screenshot from 2019-12-11 13 47 32@2x

greglobinski: I confess, I did not see the design :), so you can overlook last suggestion :)

joshwcomeau: RE: the tooltip, yeah I was initially wanting it to line up with the "V", but it was too much trouble haha. And yeah the design has it this way anyway.

As for the width, I agree it would be better to specify a `min-` and `max-` width; I just tried this and I'm running into a silly issue where it's always keeping the min-width, wrapping text:

Screenshot 2019-12-12 at 10 41 44 AM

My hunch is because this `ul` is inside the parent `li`:

Screenshot 2019-12-12 at 10 41 53 AM

Here's what it does without any sort of width constraint:

Screenshot 2019-12-12 at 10 47 35 AM

I'm actually not sure how to solve for this 🤔 other than a heavy-handed JS solution where the `<ul>` is positioned relative to the viewport instead of its parent `<li>`. Will create an issue to follow up on this, but IMO I think the fixed width is fine for now, even if it's not ideal.

joshwcomeau: https://github.com/gatsby-inc/gatsby-interface/issues/166

greglobinski: hmmmm, I will take a look

PR #164, opened 2019-12-12
Update colors token exports

greglobinski:
A quick change in exporting `colors` design-tokens, required for https://github.com/gatsby-inc/mansion/pull/5924

- stop exporting `palette`
- export `colors` from `/theme` instead of `/utils/presets`

no changes in components so I push it directly to `master` to not deal with the current stage of `dev`.

Removing  usage of `palette` in the `gatsby-interface` components is in a separate PR -> https://github.com/gatsby-inc/gatsby-interface/pull/165

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -78,7 +78,6 @@ export * from "./components/CopyButton"
 
 export { default as colors } from "./theme/colors"
 
-export { palette } from "./utils/presets/colors"
 export { fontFamilies, fontSizes } from "./utils/presets/typography"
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -76,7 +76,8 @@ export { Switch } from "./components/core/Switch"
 
 export * from "./components/CopyButton"
 
-export { colors, palette } from "./utils/presets/colors"
```
mfrachet: Won't this break every places of mansion where we import palette?

PR #174, opened 2019-12-17
Two navigation fixes

joshwcomeau:
Our navigation component needed two more tweaks:

- On mobile, the list was showing horizontally instead of vertically
- We needed to support a "split" navigation, with items on either side.

The former was a quick fix, two lines of CSS.

The latter was a bit more trouble, but I solved it by adding a new `secondaryItems` prop, wrapping both sides in a flex container, and shoving a spacer between them to force them to each end.

### Images:

With secondaryItems:

Screenshot 2019-12-17 at 4 58 18 PM

with root children (unchanged):
Screenshot 2019-12-17 at 4 58 13 PM

### DIFF
--- a/src/components/skeletons/BaseNavigation/BaseNavigation.js
+++ b/src/components/skeletons/BaseNavigation/BaseNavigation.js
@@ -14,6 +14,7 @@ const BaseNavigationContext = React.createContext()
 
 const BaseNavigation = ({
   items = [],
+  secondaryItems = [],
   children,
   isInverted = false,
   mobileNavMediaQuery = `@media (max-width: 1065px)`,
@@ -49,6 +50,7 @@ const BaseNavigation = ({
 
   const value = {
     items,
+    secondaryItems,
     rootChildren: children,
     isInverted,
     mobileNavMediaQuery,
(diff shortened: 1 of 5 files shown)


src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -138,16 +140,22 @@ BaseNavigation.Nav = ({ ...rest }) => {
...
-    <ul css={baseStyles.list.default} {...rest}>
-      {items.length > 0 &&
-        items.map(item => <Item key={item.name} item={item} />)}
-      {rootChildren && rootChildren}
-    </ul>
+    <div css={baseStyles.list.wrapper}>
+      <ul css={[baseStyles.list.side, baseStyles.list.leftSide]} {...rest}>
+        {items.length > 0 &&
+          items.map(item => <Item key={item.name} item={item} />)}
+        <div css={baseStyles.list.spacer} />
```
joshwcomeau: My initial thought was to have two separate `ul` elements with `justify-content: space-between` forcing them apart, but for some reason this caused the link items to shrink and squash, forcing them across multiple lines. Not sure why :/ but this approach seems to work great!

Elanhant: I suspect that if the link items had `word-wrap: nowrap` on them they wouldn't  `shrink and squash` 🤔 Just a guess though.

DSchau:
@Elanhant could we get this in? Would love to ship gatsby-inc/mansion#5864 sooner rather than later!

PR #176, opened 2019-12-19
chore: add branch prerelease workflow to CircleCI

Elanhant:
This PR adds a CircleCI workflow to support multiple release channels via NPM which would allow every branch have a designated channel. 

For example, if I pushed to a branch named `Elanhant/fix-some-issue`, the workflow would publish a timestamped package version `0.0.60-Elanhant-fix-some-issue-20191219T1000` assigned to a tag named `Elanhant-fix-some-issue`. Thanks to that, it is possible to install the latest version with the code from the branch using this command:
```bash
yarn add gatsby-interface@Elanhant-fix-some-issue 
```

This workflow does not use `standard-version` or git tags to keep it really simple and avoid conflicts.

`master` and `dev` branches are excluded from the workflow as they have their own release processes.


.circleci/config.yml
```diff
@@ -114,6 +126,17 @@ workflows:
   test:
     jobs:
       - is-green
+  run-branch-prerelease:
+    jobs:
+      - is-green
+      - branch-prerelease:
+          requires:
+            - is-green
```
joshwcomeau: Assuming this means that it will only publish a prerelease if the CI tests and stuff pass?

I wonder if we need this... I might have a WIP `gatsby-interface` change with some lint issues, but I still want folks to be able to test out my Mansion PR.

Plus, the CI might take some time / be sluggish, and I don't think it should depend on that.

I could be totally misunderstanding this though 😅

package.json
```diff
@@ -28,7 +28,9 @@
     "lint:all": "yarn lint src",
     "chromatic": "chromatic test",
     "scaffold:icon": "babel-node --presets @babel/env scripts/scaffold-icon/scaffold-icon.js",
-    "release": "standard-version"
+    "prepublish": "yarn build",
```
joshwcomeau: I was wondering why we didn't have this before! Glad to see it added

scripts/prerelease-branch.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * This script publishes a prerelease tag to NPM based on the current branch name
+ * It is based on this article from IBM:
+ * https://developer.ibm.com/articles/d-private-npm-modules-travis-ci-presence-insights-trs/
+ *
+ * We are bypassing standard-version here as we only want to publish a new and short-lived version to NPM
+ */
+import sh from "shelljs"
+import chalk from "chalk"
+import moment from "moment"
```
joshwcomeau: It's really not a big deal since this is a tooling-only change, but it might be good to use `date-fns`. it's a much smaller / more modular library.

Even though it doesn't matter for this change, just having `moment` as an installed dependency might prompt folks to use it in their components, and it's friggin' huge for what it does (like hundreds of kb, if I remember).

With date-fns, you could do this:

```js
import format from 'date-fns/format';

...

const version = `${pkg.version}-${normalizedBranch}-${
  format(new Date(), 'YYYYMMDD-HHmm')
}`


```

Elanhant: For sure, I don't mind replacing `moment` with `date-fns` 👍

.circleci/config.yml
Elanhant: I feel like this might need some additional discussion outside of this PR. My personal opinion is that it doesn't make much sense to publish something that is not passing tests to NPM.
As for your "lint issues" example, our `is-green` check is currently performing three actions that could fail: 
1. Running type checks for TS
1. Running tests
1. Trying to build the package
So no lint issues should prevent the check from completion.

PR #177, opened 2019-12-19
ThemeProvider

Elanhant:
This PR adds a `ThemeProvider` component based on Theme UI that allows to use theme scales with Emotion:

```javascript
import { css } from '@emotion/core'

const baseCss = theme => css(
  {
    // some static styles
    display: 'inline-block',
    verticalAlign: 'middle',
  },
  {
    backgroundColor: theme.colors.green[30],
  },
)
```
For TypeScript, we also need to explicitly set type for `theme` argument since Emotion/ThemeUI typings do not support generically typed `css` prop:
```typescript
import { Theme } from 'gatsby-interface'

function TestComponent() {
  return (
    <div
      css={(theme: Theme) => ({
        backgroundColor: theme.colors.green[30],
      })}
    >
      Lorem Ipsum
    </div>
  )
}
```

Once merged and published, this will allows us to start gradually migrating the app code to using the provider.


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export type ThemeProviderProps = {
+    children?: React.ReactNode;
+}
+
+export default function ThemeProvider({ children }: ThemeProviderProps) {
+    const theme = React.useMemo(() => getTheme(), [])
+    
+    return  (
+        <ThemeUiProvider<Theme> theme={theme}>{children}</ThemeUiProvider>
```
mfrachet: TIL; First time I see a generic inside a JSX tag  😄

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { ThemeProvider as ThemeUiProvider, useThemeUI } from 'theme-ui'
+import { getTheme, Theme } from '../../theme'
+
+export type ThemeProviderProps = {
+    children?: React.ReactNode;
+}
+
+export default function ThemeProvider({ children }: ThemeProviderProps) {
```
mfrachet: Why not to use `ThemeProvide:React.FC` in this case?

src/theme/lineHeights.ts
```diff
@@ -1,5 +1,7 @@
+import { LineHeightToken as DesignLineHeightToken } from "gatsby-design-tokens"
```
mfrachet: I think you can export from here. Instead of aliasing and export another type nope? 🤔

src/components/ThemeProvider/ThemeProvider.tsx
Elanhant: It's just a matter of preference 🙂 I favour 
```typescript
function ComponentName(props: ComponentNameProps)
```
pattern over 
```typescript
const ComponentName: React.FC<ComponentNameProps>
``` 
and, since you can't type function declarations in TS, I have to explicitly specify `children` prop type.

src/theme/lineHeights.ts
Elanhant: Babel 😞The package doesn't build if I just reexport the type here without aliasing due to Babel not being able to distinguish between a value and a type export without some heuristics. I'm actually going to add an issue to see if it is better to just use TypeScript compiler with Babel.

src/components/ThemeProvider/ThemeProvider.tsx
mfrachet: LGTM so, I wasn't sure about the implications 😊

.storybook/withTheme.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+const withTheme: StoryDecorator = (story) => {
+    return (
+        <ThemeProvider>
+            {story()}
+        </ThemeProvider>
+    )
+}
+
+export default withTheme
```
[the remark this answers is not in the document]
Elanhant: `.storybook` used to be ignored by our lint and format scripts, but it should work fine now when we have #170 merged. My changes might have predated that merge. I'm going to add an issue to run linters and formatters on all files once outstanding PRs are merged.

src/theme/colors.ts
```diff
@@ -0,0 +1,20 @@
...
+  secondaryBackground: baseColors.grey[5],
+  standardLine: baseColors.grey[20],
+}
+
+const colors: Colors = {
+  ...baseColors,
+  ...defaultColors,
+}
+
+export default colors
```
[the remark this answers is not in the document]
Elanhant: I am probably not the right person to ask questions on `gatsby-design-tokens`, perhaps @rizchelwood or @greglobinski might know more (I think #181 would be a good place to start).

For now I can say that `gatsby-design-tokens` is not intended to be used directly neither in gatsby-interface nor in the Cloud app. Instead, its tokens should be accessed via either importing code from the `theme` folder (as we do now) or by using theming (which is the intention of this PR). 
This is why I designate aliases like `baseColors` — the main "colors" scale for the consumers is going to come from `theme/index.ts`, not from the library.

src/theme/fontSizes.ts
```diff
@@ -1,6 +1,31 @@
...
+  | 11
+  | 12
+  | 13
+  | 14
+  | 15
+  | 16
+
+export type FontSizes = MapToString<TokenFontSize>
+// TODO We should probably support base font sizes other 16px
+const fontSizes = baseFontSizes.map(item => `${item / 16}rem`) as FontSizes
```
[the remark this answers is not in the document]
Elanhant: I think #181 would be a great place to discuss this!

src/theme/fonts.ts
```diff
@@ -1,12 +1,15 @@
-import { fonts as rawFonts } from "gatsby-design-tokens"
+import { fonts as rawFonts, FontToken } from "gatsby-design-tokens"
 
-const fonts = {} // Object.keys(baseFonts).map(key => baseFonts[key].join())
+export type Font = FontToken
+
+const fonts: Partial<Record<Font, string>> = {} // Object.keys(baseFonts).map(key => baseFonts[key].join())
```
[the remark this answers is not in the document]
Elanhant: Updated and added a comment explaining why we're joining stuff here :)

src/theme/radii.ts
```diff
@@ -0,0 +1,25 @@
+import { radii as baseRadii } from "gatsby-design-tokens"
+
+export type RadiusToken = 0 | 1 | 2 | 3 | 4 | 5 | 6
+
+export type Radii = [string, string, string, string, string, string, string]
+
+const radii = baseRadii.map(val =>
+  typeof val === `number` ? `${val}px` : val
```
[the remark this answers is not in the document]
Elanhant: Good call 👍 , I've added a comment which I hope is helpful

PR #179, opened 2019-12-23
Chip component

Elanhant:
Adds a `Chip` component inspired by [Material Design chips](https://material.io/components/chips/). It is going to be used for displaying CMS and code source integrations in Cloud.

### Screenshots
Screen Shot 2019-12-23 at 10 30 49 AM


src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <Chip>Lorem ipsum</Chip>
+        </div>
+      </StoryUtils.Container>
+    )
+  })
+  .add(`IconChip`, () => {
+    return (
+      <StoryUtils.Container>
+        <div>
+          <IconChip icon={<MdArrowForward />} />
```
mfrachet: I find `IconChip` and the `icon` prop are kind of redundant 😅 . It' about details and absolutely not blocking, just noticing it 😊

src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,31 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import { css } from "@emotion/core"
+import colors from "../../theme/colors"
+import fontSizes from "../../theme/fontSizes"
+
+const HEIGHT = `1.75rem`
```
mfrachet: WYT of renaming it to `SIZE` since we're relying on this to define a `minWidth` some line below? 😄

src/components/Chip/IconChip.tsx
```diff
@@ -0,0 +1,33 @@
...
+export type IconChipProps = ChipProps & {
+  icon: React.ReactNode | null;
+}
+
+export default function IconChip({ icon, children, ...props }: IconChipProps) {
+  return (
+    <Chip css={!children && iconOnlyCss} {...props}>
+      {icon && <span css={[iconCss, children && iconOffsetCss]}>{icon}</span>}
+      {children}
+    </Chip>
```
mfrachet: This components really looks like the `Chip` component with only one (or two?) conditions. Maybe I'm missing something, but can't we merge both of them? This will get rid off the duplication on the `Icon` notion (prop + component name) ???

mfrachet:
I love the shape and look of the components, this is really beautiful! 😄

src/components/Chip/Chip.stories.tsx
Elanhant: True, I guess I wanted to preserve a "base" variant for a chip. I'm going to rename `Chip` to `BaseChip` and `IconChip` to `Chip`. That way we support `icon` prop out of the box while still allowing to leverage the stripped-down component.

PR #183, opened 2020-01-07
Styled form elements and layout primitives

greglobinski:
### Form elements

The PR continues work of #158

Screenshot from 2020-01-07 15 05 52@2x

Upadates/changes (in relation to #158)
- `Field` wrappers and `Options` container use `Stack` and `Cluster` (layout primitives) to organize `Field` inner elements  and `Field` items of group type Fields
- upgrade `Formik` to `v. 2.0.8`
- refactor the `*ConnectedField`  components to use `useFormikContext` instead of HoC `connect`
- apply a series of visual/styling updates suggested/requested in  #158 

### Stack

`Stack` is a layout primitive. It's only role is to organize its children as verticaly organized elements with a defined `gap` between them.

#### Stack props

- `gap?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | string` (default value 0)
- `align?: 'justify' | 'center' | 'left' | 'right'` (default value `left`)
- `as?: 'div' | 'span' | 'section'` (default value `div`)

If we set the `gap` value as type `number` between 0 and 15, `gap` will be based on `space` design token values.

#### getStackStyle

`getStackStyle` is a helper which let us apply `Stack` styling to any other `tag` without using the `Stack` component.

```javascript
<div css={[
  getStackStyle({ gap: 2, align: `centre`})
]}>
  <div>one</div>
  <div>two</div>
  <div>three</div>
</div>
```

Screenshot from 2020-01-07 15 18 10@2x

### Cluster

`Cluster` is a layout primitive. It's only role is toorganize its children as inline organized elements with a defined `gap` between them.

#### Cluster props

- `gap?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | string` (default value 0)
- `verticalGap?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | string` (default value the value of gap)
- `align?: 'center' | 'left' | 'right'` (default value `left`)
- `as?: 'div' | 'span' | 'section'` (default value `div`)

If we set the `gap` or `verticalGap` value as type `number` between 0 and 15, `gap` will be based on `space` design token values.

#### getClusterStyle

`getClusterStyle` is a helper which let us apply `Cluster` styling to any other `tag` without using the `Cluster` component.

```javascript
<div css={[
  getClusterStyle({ gap: 2, verticalGap: 3, align: `center`})
]}>
  <div>one</div>
  <div>two</div>
  <div>three</div>
</div>
```

Screenshot from 2020-01-07 15 18 23@2x


rollup.config.js
```diff
@@ -35,7 +35,14 @@ export default {
       extensions,
     }),
     resolve({ extensions }),
-    commonjs(),
+    commonjs({
+      namedExports: {
+        "node_modules/formik/node_modules/scheduler/index.js": [
```
Elanhant: Do we know why is this needed? If yes, then it would be great to have a comment here with the explanation.

src/components/Cluster/Cluster.helpers.tsx
```diff
@@ -0,0 +1,51 @@
...
+      justifyContent:
+        props && props.align ? translateAlignToCss(props.align) : `center`,
+      margin: `calc(${gapVal} / 2 * -1)`,
+      marginTop: `calc(${verticalGapVal ? verticalGapVal : gapVal} / 2 * -1)`,
+      marginBottom: `calc(${
+        verticalGapVal ? verticalGapVal : gapVal
+      } / 2 * -1)`,
+    },
+
+    "& > * > *": {
```
Elanhant: This kind of selectors frightens me 😄Ideally, I would probably prefer to manipulate React children in some way, e.g. render a wrapper div that can have these styles:
```jsx
function Cluster({ children, gap, verticalGap }) {
  return (
    <div>
      {React.Children.map(children, (child) => {
        return <div css={{ /* cluster item styles */}}>{React.cloneElement(child)}</div>
      })}
    </div>
}
```
Another API that comes to mind is to only expose the style getters, not a component, which would be used like this:
```jsx
import { getClusterCss } from 'gatsby-interface'

const { containerCss, itemCss } = getClusterStyles({ gap: '4rem', verticalGap: '2rem', align: 'center' })

return (
  <div css={containerCss}>
    <div css={itemCss} />
    <div css={itemCss} />
    <div css={itemCss} />
  </div>
)
```

I wonder what do you and others think about this?

src/components/Cluster/Cluster.helpers.tsx
```diff
@@ -0,0 +1,51 @@
+import { Interpolation } from "@emotion/serialize"
+import space, { SpaceToken } from "../../theme/space"
+
+export type ClusterGap = SpaceToken | string
+export type ClusterAlign = `center` | `left` | `right`
+export type StackAlignCSS = `center` | `flex-start` | `flex-end`
```
Elanhant: Should it be `StackAlignCSS` or `ClusterAlignCSS`?

src/components/Cluster/Cluster.helpers.tsx
```diff
@@ -0,0 +1,51 @@
...
+    "& > * > *": {
+      margin: `calc(${gapVal} / 2)`,
+      marginTop: `calc(${verticalGapVal ? verticalGapVal : gapVal} / 2)`,
+      marginBottom: `calc(${verticalGapVal ? verticalGapVal : gapVal} / 2)`,
+    },
+  }
+}
+
+function translateAlignToCss(val: ClusterAlign): StackAlignCSS {
+  const VAL_TO_CSSVAL: {
```
Elanhant: You can declare the type as a Record here:
```typescript
const VAL_TO_CSSVAL: Record<ClusterAlign, StackAlignCSS>
```

Also, I think we can omit the `translateAlignToCss ` function entirely and just use `VAL_TO_CSSVAL[val]` instead.

src/components/Stack/Stack.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export type GetStackStyleProps = {
+  gap?: StackGap
+  align?: StackAlign
+}
+
+export function getStackStyle(props?: GetStackStyleProps): Interpolation {
+  const { gap = 0 } = props || {}
+
+  const gapVal: string = typeof gap === `string` ? gap : space[gap]
```
Elanhant: I would probably changed to use `gap` as a fallback value if `space[gap]` is not defined:
```typescript
const gapVal: string = space[gap] !== undefined ? space[gap] : gap
```
Otherwise if we pass `gap={88}` it would result in `marginTop: undefined` which is harder to debug than `marginTop: 88` 🙂

src/components/Stack/Stack.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return {
+    display: `flex`,
+    flexDirection: `column`,
+    alignItems:
```
Elanhant: Or just 
```javascript
alignItems : translateAlignToCss(props.align || 'justify')
```

src/components/Stack/Stack.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+      props && props.align ? translateAlignToCss(props.align) : `justify`,
+
+    "& > * + *": {
+      marginTop: gapVal,
+    },
+  }
+}
+
+function translateAlignToCss(val: StackAlign): StackAlignCSS {
```
Elanhant: See my comment for the similar cluster function.

src/components/Stack/Stack.tsx
```diff
@@ -0,0 +1,34 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import React from "react"
+
+import { getStackStyle, GetStackStyleProps } from "./Stack.helpers"
+
+type PropsOf<Tag> = Tag extends keyof JSX.IntrinsicElements
```
Elanhant: Nice! One thing — could you please add a `TODO` comment to replace this once #169 is merged?

src/components/Cluster/Cluster.helpers.tsx
joshwcomeau: I'm on the same page as Aleksei - I feel like the idea of targeting children that aren't rendered by a component can lead to a lot of confusion.

I like both of Aleksei's suggestions, I think we should have both options =)

src/components/Cluster/Cluster.tsx
```diff
@@ -0,0 +1,36 @@
...
+  verticalGap = 0,
+  align = `left`,
+  ...rest
+}: ClusterProps) => {
+  const Component = as
+
+  return (
+    <Component css={getClusterStyle({ gap, verticalGap, align })} {...rest}>
+      {/* intermediary wrapper needed to remove outer margins*/}
+      <div>{children}</div>
```
joshwcomeau: Aleksei's second suggestion above would mean applying that second helper on this div, which I think would help clarity quite a bit.

src/components/Cluster/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+`getClusterStyle` is a helper which let us apply `Cluster` styling to any other `tag` without using the `Stack` component.
+
+```javascript
+<div css={[getClusterStyle({ gap: 2, verticalGap: 3, align: `centre` })]}>
+  <div>one</div>
+  <div>two</div>
+  <div>three</div>
+</div>
+```
```
joshwcomeau: Thanks for the great docs!

src/components/Stack/Stack.tsx
```diff
@@ -0,0 +1,34 @@
...
+  ...rest
+}: StackProps) => {
+  const Component = as
+
+  return (
+    <Component css={getStackStyle({ gap, align })} {...rest}>
+      {children}
+    </Component>
+  )
+}
```
joshwcomeau: It could be that there's complexity that I'm not aware of, but I'm wondering if we need two separate components for `Cluster` and `Stack`; from a consumer perspective, I think I'd prefer an API like this:

```js
<Stack direction={"vertical" | "horizontal"}>
```

I know that there are differences between them around styling (including the structure, with Clusters requiring an inner div), but I don't _think_ it would add too much complexity to combine them (even if it means Stacks having an unnecessary `<div>`).

Failing that, we could still have 2 separate components, but 1 component interface that wraps them, so that consumers can still treat it as vertical/horizontal (which I think is a more helpful mental model than block/inline).

Curious what you think, maybe there's stuff I'm not aware of?

src/components/Cluster/Cluster.helpers.tsx
mfrachet: ℹ️  _I'm not really up to date with the work going on in this PR, so it can be a bit out of scope_

The job you've made @greglobinski is really good. I really like where we're going on here. I'm just having one or two concerns about some stuff here.

Except for compound components like Tabs or Routing or stuff like that, I'm not really really really for mapping and cloning children. The following thoughts are more general than the actual implementation, but just wanted to provide some insights.

- In a more general context, **it makes you think that you can compose components** but you often end up with a component getting a prop that is not wanted (I look at you IntegrationRow and your isConnected prop). And this, even in typescript.

- It forces your children to have a special shape: what if your children is only one element and you try to map over it? (you have some extra functions from React to handle this for you, but you're not free anymore in children)

- It **hides information under the hood**, you don't totally know what is exactly going on, you get some extra information from somewhere, and it may be hard to debug or to reason about

_However_ if we want to stick with a compound approach, I suggest that we create an intermediary component that receives that prop. Just to make sure we're consistent. Something like `Cluster.Node` or `Cluster.Item`? Something like:

```jsx
export const App = () => {
  return (
    <Cluster>
      <Cluster.Node as="section">Do your stuff here</Cluster.Node>
      <Cluster.Node as="section">Do your stuff here</Cluster.Node>
    </Cluster>
  );
};
```

This way, it's kind of explicit that they are linked? + it can provide a way to get rid off the `getClusterStyles` function and only provide one way to make clusters? (less solutions, less errors?) WYT?

greglobinski:
TIL!!!

> perhaps my knowledge is outdated, but I thought that selectors are processed by CSS from right to left which makes this kind of wildcards not very performant.

@Elanhant it's not! Actually only now after some reading I get to know what that "from right to left" means :)

> I'm on the same page as Aleksei - I feel like the idea of targeting children that aren't rendered by a component can lead to a lot of confusion.

@joshwcomeau Generally yes, but I believe that not always. In case like this when the only purpose of the Component (container type) is to organize it's children there is no confusion that  it does to what it was created to. 

> Except for compound components like Tabs or Routing or stuff like that, I'm not really really really for mapping and cloning children.

@mfrachet I feel the same

To sum up, I think that we should go with the Aleksei's second variant

```
import { getClusterCss } from 'gatsby-interface'

const { containerCss, itemCss } = getClusterStyles({ gap: '4rem', verticalGap: '2rem', align: 'center' })

return (
  <div css={containerCss}>
    <div css={itemCss} />
    <div css={itemCss} />
    <div css={itemCss} />
  </div>
)
```

Especially that there is already a helper function to apply `Cluster` styles to any tag or component. https://github.com/gatsby-inc/gatsby-interface/blob/f6ccda5969e25c96c734bb87ac4e0fdf64613cdb/src/components/Cluster/Cluster.helpers.tsx#L14 Extending it to Aleksei's version will solve the  performance issue. 

What do you think about removing `Cluster` component and maintain only `getClusterStyles()` helper? (AFAIU @joshwcomeau already voted for maintain the two ways Component and getter)

src/components/Stack/Stack.tsx
greglobinski: Besides the difference you described already. 

I believe that they are two different types of organizing elements. The Cluster wraps elements if there is no place in one row, it organizes `inline` type elements so there is no `justify` alignment. 

If we would need a `horizontal` Stack, which we do not need now (for the forms), but probably will in the future, the elements  which do not fit in one line should be managed by `overflow: hidden` settings.

PS:

> even if it means Stacks having an unnecessary <div>

That's really surprising :) (hearing the above from you)

src/components/Stack/Stack.helpers.tsx
greglobinski: Absolutely!

greglobinski:
> The doc doesn't really cover how to "retroactively" split PRs up, and I don't know to what extent we want to enforce it (that's @Elanhant's call!), but IMO it would be worth taking the time to split this into at least 2 PRs, and possibly more (depending on how similar all of the form updates are). I'd be happy to jump in a call and pair on it!

@joshwcomeau I can do that, let say

PRs: 
1. layout components/styles (because form elements depend on them)
2. standard element components > `*Field`
3. block type components > `*FieldBlock`
4. connected field components > `*ConnetedField`

cc/ @Elanhant ?

greglobinski:
closed in favor of https://github.com/gatsby-inc/gatsby-interface/pull/190

Issue #166, opened 2019-12-12
Dynamic-width nav dropdowns


joshwcomeau:
With the new version of the navigation, we're enforcing that all navigation dropdowns are 450px wide.

A more reasonable approach would be to set a min- and max-width. Unfortunately, we wind up with a text-wrapping issue:

<img width="297" alt="Screenshot 2019-12-12 at 10 41 44 AM" src="https://user-images.githubusercontent.com/6692932/70727168-36b61f00-1ccd-11ea-90f6-fc2eeb93bf23.png">

This is because the dropdown `<ul>` is positioned absolutely within a relative `<li>`:

<img width="463" alt="Screenshot 2019-12-12 at 10 41 53 AM" src="https://user-images.githubusercontent.com/6692932/70727197-4170b400-1ccd-11ea-997e-ae859ae8366f.png">

By default, it tries to match the same width. Here's how it looks with no min/max width, and no enforced width:

<img width="251" alt="Screenshot 2019-12-12 at 10 47 35 AM" src="https://user-images.githubusercontent.com/6692932/70727225-4e8da300-1ccd-11ea-82b5-28c82218535f.png">

We should find a workaround so that `<ul>`s aren't constrained in this way.

PR #189, opened 2020-01-10
Navigation: dynamic-width, centered dropdowns + tab navigation improvements

greglobinski:
### 1. Dynamic-width for Dropdowns

- resolves: https://github.com/gatsby-inc/gatsby-interface/issues/166
- `Dropdown` container `width` adapts to its content, no more hard-coded `450px` 

Before:

Navigation - default example  Storybook (3)

After:

Navigation - default example  Storybook (6)

### 2. Dropdown positioning and viewport colision detection

- `Dropdown` container and its beak are centered in relation to its parent `Item` link
- When `Dropdown` at its default positioning does not fit into `viewport` it is shifted, but the beak stays centered in relation to `Item` link as usual
  
Before: 

Navigation - default example  Storybook (4)

After: 

Navigation - default example  Storybook - Edited (1)

### 3. Tab navigation improvements
 
- when `focus` leaves `Dropdown` component, it closes itself
- `ToggleDropdown` rotates indicating the state of `Dropdown`

Before: 

Navigation - default example  Storybook - Edited

After 

Navigation - default example  Storybook (7)

## TESTING

Please test it in Safari, I did it by BrowserStack, generally it works, but it renders strange lines around the `shifted` dropdowns and I was not able to verify if they are really produced by Safari or BrowserStack.


src/theme/styles/navigation.js
```diff
@@ -84,31 +85,46 @@ styles.List = {
...
     height: 12,
     content: `" "`,
-    transform: `rotate(45deg)`,
     borderRadius: `2 0 0 0`,
     background: colors.white,
     boxShadow: `-3px -3px 10px ${hexToRGBA(colors.lilac, 0.1)}`,
-    willChange: `transform`,
-    transitionProperty: `transform`,
-    transitionDuration: transition.speed.default,
+    transform: `translateX(calc(-50% + (${offset} * -1))) rotate(45deg)`,
```
joshwcomeau: Might be good to add a comment explaining this - it's a pretty dense line.

src/utils/helpers/callIfTrueInNextLoop.js
```diff
@@ -0,0 +1,7 @@
+export function callIfTrueInNextLoop(condition, callback) {
```
joshwcomeau: Nit: the word "loop" here initially confused me. Maybe "callIfTrueInNextEventLoop" might be a bit clearer?

Elanhant:
> * The position starts out wrong (only in Safari, not Chrome). Guessing this has to do with rendering differences... maybe wrapping the updates in a `requestAnimationFrame` could help?

I have a feeling that this might be related to using `useEffect` instead of `useLayoutEffect` to decided when to call `setDropdownOffsets`. According to [React docs](https://reactjs.org/docs/hooks-reference.html#uselayouteffect):
> The signature is identical to useEffect, but it fires synchronously after all DOM mutations. Use this to read layout from the DOM and synchronously re-render. Updates scheduled inside useLayoutEffect will be flushed synchronously, before the browser has a chance to paint.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -235,49 +258,117 @@ BaseNavigation.DropdownToggle = ({
...
+    const VIEWPORT_FIT_MARGIN = 20
+
+    React.useEffect(() => {
+      if (dropdownRef.current) {
+        const windowWidth =
+          window.innerWidth || document.documentElement.clientWidth
+
+        dropdownRef.current.style.visibility = `hidden`
+        dropdownRef.current.style.display = `block`
+        const { left, right } = dropdownRef.current.getBoundingClientRect()
```
Elanhant: Nitpicking: this line looks kinda intrusive among `dropdownRef.current.style`, it's better to move it below, closer to `leftFit` calcualtion 🙂

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -235,49 +258,117 @@ BaseNavigation.DropdownToggle = ({
...
+    }
+
+    return (
+      <div ref={dropdownRef} css={baseStyles.dropdown} {...rest}>
+        <ul
+          // id to associate with aria-controls on BaseNavigation.Item
+          id={`${item.name}-dropdown`}
+          onKeyDown={e => {
+            // handle closing dropdown on `esc`
+            if (e.keyCode === 27) {
```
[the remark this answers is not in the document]
Elanhant: I remember it being quite a confusing thing — some IDEs like WebStorm say that `e.keyCode` is deprecated and `.code` should be used instead, and yet [Edge 18 is not supporting that](https://caniuse.com/#feat=keyboardevent-code) 🤷‍♂ If we don't care about that, then I'm OK with switching to `.code`.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -235,49 +258,117 @@ BaseNavigation.DropdownToggle = ({
...
+              toggleDropdown(false)
+            }
+            return
+          }}
+        >
+          {dropdownItems.length > 0 &&
```
Elanhant: We don't really need to check for `dropdownItems.length > 0` since `.map` over an empty array will return an empty array and React will not render anything.

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -235,49 +258,117 @@ BaseNavigation.DropdownToggle = ({
...
+  isLast,
+  onLastItemBlur,
   ...rest
 }) => {
-  const {
-    components: { DropdownItem },
-  } = BaseNavigation.useNavigationContext()
+  const linkRef = React.useRef()
+
+  if (isLast) {
```
Elanhant: I'm afraid we can't use hooks inside conditions or loops. If you only want to call `onLastItemBlur` for a specific item, you'll have to add your own listener, e.g.
```javascript
const linkRef = React.useRef()

const onBlurListener = (e) => {
  if (isLast && onLastItemBlur) {
    onLastItemBlur(e)
  }
}

useEventListener(`blur`, onLastItemBlur, linkRef)
```

Elanhant: Also, why don't we just pass that callback to `<Link/>`? Looking at that component's type definitions, it should support `onBlur`:
```jsx
<Link onBlur={isLast ? onLastItemBlur : undefined} activeClassName="nav-item-active" to={linkTo}>
  {name}
</Link>
```

src/components/Navigation/Navigation.js
```diff
@@ -109,12 +109,21 @@ Navigation.ItemLink = delegated => {
...
 
 Navigation.Dropdown = delegated => {
-  const { mobileNavMediaQuery } = BaseNavigation.useNavigationContext()
+  const {
+    mobileNavMediaQuery,
+    dropdownOffsets,
+  } = BaseNavigation.useNavigationContext()
+
+  const offset =
+    dropdownOffsets[delegated.item.name] &&
```
Elanhant: Let's make it a bit clearer by adding an additional variable:
```javascript
const itemOffsets = dropdownOffsets[delegated.item.name]
const offset = itemOffsets && itemOffsets.offset
  ? `${itemOffsets.offset}px`
  : `0px`
```

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -165,10 +171,13 @@ BaseNavigation.Item = ({ item, children, ...rest }) => {
   const dropdownChildren = children || false
   const itemHasDropdown = dropdownItems.length > 0 || dropdownChildren
 
-  const ref = React.useRef()
+  const itemRef = React.useRef()
+
+  useEventListener(`mouseenter`, () => toggleDropdown(true), itemRef)
+  useEventListener(`mouseleave`, () => toggleDropdown(false), itemRef)
```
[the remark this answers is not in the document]
greglobinski: No special reason, 
- personal preference to use `addEventListener` instead of `onClick`
- using `hooks` for 'not to repeat yourself'

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -217,16 +226,30 @@ BaseNavigation.DropdownToggle = ({
...
+  useEventListener(`blur`, () => onToggleBlur(false), toggleRef)
+
+  function onToggleBlur() {
+    // close Dropdown if next activeElement is ItemLink
+    callIfTrueInNextLoop(
+      () =>
+        toggleRef.current.parentElement ===
+        document.activeElement.parentElement,
+      () => toggleDropdown(false)
+    )
```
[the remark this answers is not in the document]
greglobinski: I agree, that keeping track of dropdowns states in Navigation context would be more reliable solution. Although I planned to leave this refactoring to the next iteration, but maybe I try to make it now.

src/components/skeletons/BaseNavigation/BaseNavigation.js
greglobinski: Good point!

PR #193, opened 2020-01-15
styled form elements: *FieldBlock components 

greglobinski:
### Form elements as Block components

The PR contains an unchanged part of  #183  
and is an update to #190


src/components/form/components/CheckboxFieldBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+    label,
+    labelSize,
+    error,
+    hint,
+    className,
+    validationMode,
+    ...rest
+  } = props
+
+  const isRequired = rest.required ? rest.required : false
```
mfrachet: I don't think it's necessary, since required will be true or undefined I guess. If you need to keep the boolean type, you can cast the rest.required thing instead 😊 

```
const isRequired = Boolean(res.required) // true when true and false when undefined
```

src/components/form/components/CheckboxFieldBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+  labelSize?: FormFieldLabelSize
+  error?: ReactNode
+  hint?: ReactNode
+  validationMode?: ErrorValidationMode
+} & CheckboxFieldControlProps
+
+const CheckboxFieldBlock = React.forwardRef<
+  HTMLInputElement,
+  CheckboxFieldBlockProps
+>((props, ref) => {
```
mfrachet: 👍 forwardRef + props spreading makes external libraries so much useful! Thanks for this 💪

src/components/form/components/CheckboxGroupFieldBlock.tsx
```diff
@@ -0,0 +1,72 @@
...
+    options,
+    error,
+    hint,
+    className,
+    validationMode,
+    value: fieldValue,
+    ...rest
+  } = props
+
+  const isRequired = rest.required ? rest.required : false
```
mfrachet: Same as before (since they exist in multiple places, I won't flood the PR with these comments 😄 , last time you'll read about this 😎 )

src/components/form/components/CheckboxGroupFieldBlock.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <CheckboxGroupField.Option
+              value={value}
+              checked={fieldValue.includes(value)}
+              {...rest}
+            />
+            <CheckboxGroupField.OptionLabel optionValue={value}>
+              {label}
+            </CheckboxGroupField.OptionLabel>
+          </CheckboxGroupField.OptionWrapper>
+        ))}
```
mfrachet: _This comment has value only if we leave people use these forms and for a usage outside .com_

Data oriented components vs composition. It doesn't really matter if it's for our own usage, but if this is used by other folks (for themes?), I think the composition approach to take over the data oriented thing. It will leave people freedom about the way they want to build the ui.

But again, I don't really get for now what will be the goal of all these component: if there are here for .com or for an external usage :)

src/components/form/stories/form.CheckboxField.stories.tsx
```diff
@@ -54,6 +55,36 @@ storiesOf(`form`, module)
...
             </CheckboxField.Wrapper>
           </CheckboxField>
+
+          <CheckboxFieldBlock
+            id="input-example-1b"
+            label="Builds enabled"
+            labelSize={size}
+            onChange={e => action(`Change`)(e.target.value)}
+            error={error}
+            hint={hint ? hint : `This one is built with 'CheckboxFieldBlock'`}
```
mfrachet: NIT: Just a simple trick to avoid ternaries based on the same "host" value:

```
 hint={hint || `This one is built with 'CheckboxFieldBlock'`}
```

src/components/form/stories/form.CheckboxGroupField.stories.tsx
```diff
@@ -40,13 +42,39 @@ storiesOf(`form`, module)
...
       <StoryUtils.Container>
         <Wrapper>
+          <CheckboxGroupFieldBlock
+            id="example-1a"
+            label="Fieldset Legend"
+            options={options}
+            error={error}
+            onChange={e => {
+              const target = e.currentTarget
+              const valueArray: string[] = [...fieldValue] || []
```
mfrachet: I think you can also remove the right side of the `||` . On the left side, you're already creating an array and filling with some values. The array do exist (but maybe not the values)

src/components/form/components/CheckboxGroupFieldBlock.tsx
greglobinski: They are for the Cloud, however I believe that in next iteration they could be pretty much customizable  with design tokens provided by `theme`, but that needs additional work. The "composition approach" fulfill 'standard' `*Fields` components https://github.com/gatsby-inc/gatsby-interface/pull/192

src/components/form/stories/form.CheckboxField.stories.tsx
greglobinski: nice!

PR #194, opened 2020-01-16
styled form elements: *ConnectedField components

greglobinski:
### Form elements as Connected components (with Formik)

The PR contains an unchanged part of  #183  
and is an update to #190 

Screenshot from 2020-01-16 01 09 23@2x


src/components/form/components/CheckboxConnectedField.tsx
```diff
@@ -0,0 +1,41 @@
...
+  } = useFormikContext()
+  const error = getIn(errors, props.name)
+  const isTouched = getIn(touched, props.name)
+  const value = getIn(values, props.name)
+
+  return (
+    <CheckboxFieldBlock
+      id={id}
+      label={label}
+      error={isTouched && error && error}
```
mfrachet: Duplication on error here! 😊

src/components/form/components/CheckboxGroupConnectedField.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <CheckboxGroupFieldBlock
+      id={id}
+      label={label}
+      error={isTouched && error && error}
+      value={value}
+      onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+        const target = e.currentTarget
+        const valueArray = [...value] || []
```
mfrachet: I'm not sure the safety check is necessary here since you already safety check on line 30 and that you, at worse, fill an array with undefined value on the left side of the operator

greglobinski:
> I'm just wondering, because I don't have a more long term vision on the future project. Will these be useful for themes? Or is it something that we'll put in .com? I'm not sure to get where they will be used

@mfrachet They are for the Cloud, however... as @Elanhant said

> As for the lib vs app concern, I initially planned to keep it in the app. However, we might benefit from having it in gatsby-interface since we'd have type definitions for every component which, if I remember correctly, can be used by IntelliSense even when using these components in vanilla JS.

https://github.com/gatsby-inc/gatsby-interface/pull/158#issuecomment-562665415

PR #199, opened 2020-02-06
feat: fit fallback text to Avatar size

### LINKED ISSUE #136: Avatar "placeholder" text should scale to fit its container
Currently we have font size fixed for the text displayed in "placeholder" avatar:
Screen Shot 2019-11-25 at 10 19 32 AM
While this works for cases like "+1", "+2" etc, it might get out of hand if we have "+17" or something like that, especially for smaller Avatar sizes.

Elanhant:
Fixes #136 by using binary search to find the most fitting font size for an avatar fallback text. Inspired by and based on [textFit](https://github.com/STRML/textFit).

ezgif-3-7cc8229a2654

### DIFF
--- a/__tests__/__snapshots__/Storyshots.test.js.snap
+++ b/__tests__/__snapshots__/Storyshots.test.js.snap
@@ -302,6 +302,7 @@ exports[`Storyshots Avatar AvatarsGroup 1`] = `
               >
                 <span
                   aria-label="John Doe"
+                  style="font-size: 0.6875px;"
                 >
                   JD
                 </span>
@@ -323,6 +324,7 @@ exports[`Storyshots Avatar AvatarsGroup 1`] = `
               >
                 <span
                   aria-label="More users not displayed"
+                  style="font-size: 0.6875px;"
                 >
                   +1
                 </span>
--- a/__tests__/__snapshots__/Storyshots.test.js.snap
+++ b/__tests__/__snapshots__/Storyshots.test.js.snap
@@ -153,6 +153,259 @@ exports[`Storyshots Avatar Avatar 1`] = `
 </div>
 `;
 
+exports[`Storyshots Avatar Avatar fallback text fit 1`] = `
+.emotion-6 {
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  min-height: 100vh;
+  -webkit-box-pack: center;
+  -webkit-justify-content: center;
+  -ms-flex-pack: center;
+  justify-content: center;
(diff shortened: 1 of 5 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -43,7 +61,9 @@ export default function Avatar({
       {src ? (
         <img css={imageCss} src={src} alt={label} />
       ) : (
-        <span aria-label={label}>{fallback}</span>
+        <span aria-label={label} ref={textFitter}>
```
mfrachet: ❓ this is the first time I'm seeing a reference passing without using `React.createRef`, `useRef` or the callback assignement.

What I think got me in trouble is the lack of the `.current` property on the ref in `textFitter` returning function.

Just to make sure I'm understanding what's going on here 😅

__tests__/__snapshots__/Storyshots.test.js.snap
```diff
@@ -302,6 +302,7 @@ exports[`Storyshots Avatar AvatarsGroup 1`] = `
               >
                 <span
                   aria-label="John Doe"
+                  style="font-size: 0.6875px;"
```
mfrachet: Maybe we can add more snapshots for different sizes? just to make sure we don't break the algorithm you've written in `fitText` 😊

src/components/Avatar/Avatar.helpers.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const maxWidth = parent.clientWidth * maxTextWidth
+    const maxHeight = parent.clientHeight * maxWidth
+
+    let fontSize = low
+    let mid
+
+    // Binary search to find the most fitting font size
+    while (low <= high) {
+      mid = (high + low) / 2
+      element.style.fontSize = mid + "px"
```
mfrachet: I'm curious why you're changing the element style fontSize here since you already do this in line 52 (2 reflows?)

Elanhant: This is essential for the binary search to work — at every `while` block execution we are trying a new font size candidate on the text and then compare its width and height to max allowed values. Sooner or later we're going to settle on one most fitting size, which we are going to finally apply at line 52.

Elanhant: P.S. love your new user pic 💜😄

__tests__/__snapshots__/Storyshots.test.js.snap
Elanhant: This is actually quite interesting — apparently, base font size for a snapshot is `1` 😆 
Screen Shot 2020-02-06 at 10 12 54 PM

Elanhant: Done, I've added a new story to cover text fit for different avatar sizes, and it was captured in a snapshot. We still have to remember about the base font size though :)

PR #200, opened 2020-02-08
chore: remove emotion serializer from jest snapshots

Elanhant:
No more Emotion-generated CSS rules in snapshots; hopefully it will help make snapshots actually useful.


Elanhant:
> Hm, I'm not sure I understand: don't we want to be alerted if a style changes after we change some code?

With our current state of things in this repo style snapshots became more of a burden than an indicator that something was incorrectly changed. We had a discussion last week with @mfrachet, @greglobinski and @amberleyromo if we feel that we should drop these snapshots for now. 

Instead we're going to research tools for visual regression testing.

Elanhant:
> But are snapshot tests shallow? which is to say, when I run a snapshot test on `OtherLayout`, does it actually render all the children components? Since I don't think I've ever _actually_ had my scenario happen, and it seems like it should have by now.
> 

Our snapshots shouldn't be shallow — the entire React DOM tree gets rendered as string both for unit tests and Storybook stories.

mfrachet:
Snapshot testing is hard because they create a lot of noise and everybody is not testing the same "thing" while snapshotting a component. From my point of view, it's super helpful to keep them for UI components when using css-in-js tools. This helps prevent style modifications and stuff like that as @joshwcomeau as mentioned.

_However_ 

Since we are thinking of chromatic as an alternative for our UI regressions, taking care of snapshots for CSS don't look necessary anymore, at least for me.

The good point is that we are now able to make "simple" snapshots tests on _what_ the component _should_ do (rendering an error? log? content? in which cases?) instead of having a lot of pollution here 😄  (what @Elanhant  has in mind, I think :) )

Elanhant:
Merging this PR; visual regressions are going to be traced with [Chromatic](https://www.chromaticqa.com/).

PR #201, opened 2020-02-08
feat: Breadcrumb component in TypeScript

Elanhant:
Updates `Breadcrumb` component to use TypeScript; also adds `<ol>` tag to adhere with A11Y recommendations (https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html)


src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Link } from "gatsby"
+import { PropsOf } from "../../utils/types"
+import { ThemeCss } from "../../theme"
+
+export type BreadcrumbProps = Omit<JSX.IntrinsicElements["nav"], "ref">
+
+export default function Breadcrumb({ children, ...rest }: BreadcrumbProps) {
+  return (
+    <nav aria-label="breadcrumb" {...rest}>
+      <ol
```
joshwcomeau: TIL that `ol` is useful for breadcrumbs! Makes sense, since it is ordered.

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <svg
+      width="6"
+      height="9"
+      viewBox="0 0 6 9"
+      fill="none"
+      css={breadcrumbArrowCss}
+      {...props}
+    >
+      <path d="M6 4.5L1.5 8.39711L1.5 0.602886L6 4.5Z" fill="#7F7C82" />
```
joshwcomeau: Hm, is `#7F7C82` a grey we already have in our design tokens? If so it might be good to use it here

mfrachet: I didn't find it in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-design-tokens/src/colors.js :/

Elanhant: I will check if we can just replace it with a token color; we should strive to 0% non-standard colors :)

PR #204, opened 2020-02-17
ToggleSwitch and ToggleCheckbox components

Elanhant:
This PR introduces two new components as a replacement for `Switch` and `Toggle`:
* `ToggleCheckbox` (replaces `Toggle`) — basically just a glorified checkbox to be used in any "on/off" scenarios: 
  Screen Shot 2020-02-17 at 9 17 40 AM
* `ToggleSwitch` (replaces `Switch`) — looks similar to a toggle, but it has *two* labels — one for the "on" state and another for "off": 
  Screen Shot 2020-02-17 at 9 17 57 AM
  

While these components look similar, they function quite differently: `ToggleCheckbox` works as a **checkbox**, while `ToggleSwitch` works as a **radio button**. This is based on what kind of value they represent:
* `ToggleCheckbox`, as any other checkbox, represents a *boolean* value ("on/off", "true/false", "confirm/do not confirm")
* `ToggleSwitch`, however, represents a choice between *string* values that might not be easily mapped to a boolean value. For example, we are currently using `Switch` in Cloud as a way for users to choose their billing period — monthly or yearly. Is monthly "true" and yearly "false", or is it vice versa? If we were to use a one single checkbox (like we do currently), how do we convey that there are actually two different options to screen readers? For these reasons, `ToggleSwitch` is using radio inputs under the hood.

P.S. the old `Switch` and `Toggle` components are not removed by this PR, this will be done once Cloud is using the new components.


src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,341 @@
...
+    )
+  })
+  .add(`before or after label`, () => {
+    const [subInterval, setSubInterval] = useState(true)
+    const [renew, setRenew] = useState(true)
+
+    return (
+      <React.Fragment>
+        {/* 
+        // @ts-ignore */}
```
mfrachet: Why is this necessary?

src/components/Toggle/ToggleCheckbox.tsx
```diff
@@ -0,0 +1,88 @@
...
+export type ToggleCheckboxProps = Omit<
+  BaseToggleProps,
+  "label" | "ref" | "id"
+> & {
+  id: string
+  label: React.ReactNode
+  labelPosition?: "start" | "end"
+}
+
+export default React.forwardRef<HTMLInputElement, ToggleCheckboxProps>(
```
mfrachet: 👍 for forwarding ref in every PR you create, this is super nice

src/components/Toggle/ToggleGutter.tsx
```diff
@@ -0,0 +1,65 @@
...
+              width: `24px`,
+            },
+          },
+        },
+      ]}
+    />
+  )
+}
+
+ToggleGutter.tagName = "span"
```
mfrachet: Is this necessary for testing or anything particular? I mean something like the `displayName` trick?

src/stylesheets/a11y.ts
```diff
@@ -0,0 +1,12 @@
+import { Interpolation } from "@emotion/core"
+
+export const visuallyHiddenCss: Interpolation = {
```
mfrachet: Thank you for this, we need this and we need to use it in .com also 😊

src/components/Toggle/ToggleGutter.tsx
```diff
@@ -0,0 +1,65 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import { AtomTone } from "../../theme/types"
+import { Theme } from "../../theme"
+
+export type ToggleGutterProps = JSX.IntrinsicElements["span"]
+
+export default function ToggleGutter(props: ToggleGutterProps) {
```
greglobinski: Don't we need an `onClick` method on this, users expect that clicking the input will toggle it's value, don't they?

src/components/Toggle/ToggleGutter.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <span
+      {...props}
+      aria-hidden
+      css={(theme: Theme) => [
```
greglobinski: We need to set a `focus` style for this, now the focus outline has non proper blue color

src/components/Toggle/ToggleSwitch.tsx
```diff
@@ -0,0 +1,87 @@
...
+  valueOn: string
+  valueOff: string
+  labelOn: React.ReactNode
+  labelOff: React.ReactNode
+  tone?: AtomTone
+  className?: string
+  style?: React.CSSProperties
+}
+
+export default function ToggleSwitch({
```
greglobinski: I think we should override default `keypress` event listeners for the whole component. When I see `Toggle` element I expect that it will work as `checkbox`, I can change its value value with `space` key, but because there are, under the hood, `radio` buttons we have to use `arrow keys`, and that's not intuitive

src/components/Toggle/Toggle.stories.tsx
Elanhant: This comment is used in stories for the old `Toggle` component which is written in JS, not TS. Since I'm going to remove it completely, I decided not to waste time trying to fix some TS errors related to it :)

src/components/Toggle/ToggleCheckbox.tsx
Elanhant: Can't wait when components will be forwarding refs by default!

src/components/Toggle/ToggleGutter.tsx
Elanhant: It is used as part of CSS selectors in `ToggleCheckbox` and `ToggleSwitch` like this:
```jsx
<div
  css={theme => ({
    [`&:focus-within > ${ToggleGutter.tagName}`]: ToggleGutter.getFocusCss(
      theme
    ),
  })}
>
```
It more clearly conveys the intention of targeting a specific element and not just `span` in general.

src/components/Toggle/ToggleGutter.tsx
Elanhant: The `ToggleGutter` itself is not focusable, applying pseudo-focused styles is taken care of by `ToggleSwitch` and `ToggleCheckbox` via using `ToggleGutter.getFocusCss`.

As for the outline color, the original color was too light so I switched to a blue color, but I'm wondering what does @fk think?
Screen Shot 2020-02-20 at 9 14 06 AM
Screen Shot 2020-02-20 at 9 14 11 AM

src/components/Toggle/ToggleSwitch.tsx
Elanhant: Good point! This is done now.

Elanhant:
I've updated the PR to include the following:
* `ToggleSwitch` supporting click-to-toggle behaviour — it is now possible to click on a label/toggle itself to change the value to the opposite one
* Stories for both `ToggleSwitch` and `ToggleCheckbox` are rewritten with [Component Story Format (CSF)](https://storybook.js.org/docs/formats/component-story-format/) and follow the guidelines listed in #205 
* Some unit tests for `ToggleSwitch` and `ToggleCheckbox` have been added.

.storybook/config.js
```diff
@@ -96,6 +96,8 @@ addDecorator((storyFn, context) => withConsole()(storyFn)(context))
 
 addDecorator(withA11y)
 
+// Storybook now has first-class support for hooks, but Storyshots for some reason still fails
+// so we only transform stories to React elements in test environment
```
mfrachet: Yep, it's because we were breaking the rule of hook while doing `story()` which is a function and not a react "thing".

We have to trick the way storyboko renders to take benefit from hooks. OR we can wrap our stories in a component so that the story returns a function (the wrapper component) which will display the actual React stuff 😊

mfrachet:
Looks like a jest config has been broken in CI :/

Elanhant:
Going to merge this as the tests are fixed and comments regarding functionality have been addressed.

PR #206, opened 2020-02-17
Visual regression testing

Elanhant:
This PR adds two new commands:
* `yarn test:a11y` — runs [Axe](https://www.deque.com/axe/) accessibility checks for Storybook stories
* `yarn test:visual-regression` — performs visual regression testing for Storybook stories

`yarn test` will also run visual regression testing but it **will not** run a11y tests since there are about 25 failures already that we need to address before enabling `test:a11y` to our CI 🙂 

Both new commands rely on [StoryShots addon](https://github.com/storybookjs/storybook/tree/master/addons/storyshots/storyshots-core) as well as its [puppeteer addon](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer). 

I've also fixed some a11y issues in some of the stories.

PR #207, opened 2020-02-21
chore: setup Chromatic in CircleCI

Elanhant:
This PR sets up CI integration with Chromatic for visual regression testing based on Storybook stories:
1. CircleCI will run `yarn chromatic:ci` to trigger Chromatic tests
1. If there are any "noticeable" differences compared to a base snapshot of a story, Chromatic will ask to review the changes; PR checks are going to fail in this case.
1. If the changes are approved, the CircleCI job must be restarted manually.

If you are a collaborator, you should be able to login to Chromatic with your GitHub account and review regressions. If you encounter any problems with that, ping me.

Some additional changes from this PR:
* Stop ESLint from ignoring `. storybook` folder — I have no idea why, but currently ESLint is ignoring any files located in `.storybook`, which makes it difficult to commit changes to those files.
* `storybook-static` is added to `.gitignore` — there's no point in keeping static build of Storybook in the repo; our hosting setup with Netlify is running `build-storybook` anyway.
* Some stories had to be updated to stop non-deterministic behaviour (stories for `Avatar` relied on cat image randomizer) or excluded from Chromatic tests completely (`DecorativeDots`); otherwise each Chromatic run would detect a new regression.


src/components/DecorativeDots/DecorativeDots.stories.tsx
```diff
@@ -16,6 +16,8 @@ storiesOf(`DecorativeDots`, module)
       sidebar: README,
       includePropTables: [DecorativeDots],
     },
+    // TODO leverage isChromatic and testSafeSample/testSafeMathRandom instead of ignoring the story
```
joshwcomeau: Cool to hear there are ways to test "impure" stories.

I think a while ago, I suggested 'cheating' when `process.env.NODE_ENV` was `test`, we could always render a fixed set of circles in this place. But yeah, if there's a way to use Chromatic utilities, that's even better!

Elanhant: And you actually do use `process.env.NODE_ENV` check in those two functions 🙂. However, seems like its value is different for Chromatic runs, so we can either try to use some other env variable or update `DecorativeDots` to support a prop to make it deterministic.

mfrachet:
I'm not really used to chromatic, never played with it or any visual regression testing tools. I've run it locally and it has some diffs between before and after. Do we have the same problem as https://github.com/gatsby-inc/gatsby-interface/pull/206 ? Or maybe it's not meant to be run locally?

Screenshot 2020-02-24 at 09 19 27

Build 16 in the Chromatic UI

Elanhant:
> I'm not really used to chromatic, never played with it or any visual regression testing tools. I've run it locally and it has some diffs between before and after. Do we have the same problem as #206 ? Or maybe it's not meant to be run locally?
> 
> Screenshot 2020-02-24 at 09 19 27
> 
> Build 16 in the Chromatic UI

Ok so it turns out the problem with local runs is the Futura fonts that are only available locally and get included in the static Storybook build which is created when you run `yarn chromatic`. I'll see if I can exclude these fonts from Chromatic tests.

mfrachet:
Well done @Elanhant , working on my side 😊

PR #209, opened 2020-02-24
chore: setup accessibility tests with axe

Elanhant:
This PR extracts a11y storyshots tests from #206.

A new script, `yarn test:a11y`, runs [Axe](https://www.deque.com/axe/) accessibility checks on every story; it requires Storybook to be run locally (or built via `yarn build-storybook`, but you'll have to provide `CI` environment variable — `CI=1 yarn test:a11y`).

This PR also fixes some a11y issues with the current stories and components, but there are still 24 unaddressed errors 🙂 Because of that, this script is not yet added to CI steps.

I've also removed the previous storyshots tests since they are replaced by Chromatic.


jest.config.js
```diff
@@ -12,7 +12,14 @@ module.exports = {
     ".+\\.(css|styl|less|sass|scss)$": `identity-obj-proxy`,
     ".+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|md)$": `<rootDir>/__mocks__/fileMock.js`,
   },
-  testPathIgnorePatterns: [`node_modules`, `.cache`, `cypress`],
+  testPathIgnorePatterns: [
+    `node_modules`,
+    `.cache`,
+    `cypress`,
```
mfrachet: NIT but we don't have cypress here, so maybe we can remove it?

package.json
```diff
@@ -12,6 +12,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "test": "jest",
+    "test:a11y": "jest __tests__/AccessiblityStoryshots.test.js --testPathIgnorePatterns=[]",
```
mfrachet: ❓ why do we need the testPathIgnorePatterns here? We're only testing the `AccessiblityStoryshots` spec file nope?

Elanhant: It is to override `testPathIgnorePatterns` specified in `jest.config.json`, otherwise this script would not run any tests (since the a11y storyshots are ignored in the config).

In moments like this I am especially bitter about JSON not supporting comments 😄

PR #210, opened 2020-02-25
Add support for external links

joshwcomeau:
For the Willit.build project, we need a link to Gatsby Cloud. Since WIB is its own project, this needs to be an external link:

Screenshot 2020-02-25 at 11 21 28 AM

This component uses Navigation.Button, which is ultimately a wrapper around BaseLink.

In my side-projects, I often add a bit of logic to my base link component, to switch to an `<a>` if the link appears to be external. 

Alternatively I could add a prop, `isExternal`, but in my experience it can be deciphered correctly almost all of the time, and we can always add an override if we run into an edge-case.

### DIFF
--- a/__tests__/__snapshots__/Storyshots.test.js.snap
+++ b/__tests__/__snapshots__/Storyshots.test.js.snap
@@ -360,6 +360,32 @@ exports[`Storyshots BaseLink default 1`] = `
 </div>
 `;
 
+exports[`Storyshots BaseLink external 1`] = `
+<div>
+  <div
+    class="storybook-readme-story"
+  >
+    <div>
+      <div>
+        <div
+          class="css-3ga7ji"
+        >
+          <div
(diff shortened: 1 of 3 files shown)


joshwcomeau:
> Can we instead change the Navigation and BaseNavigation components? We already have AnchorButton as a non-Gatsby-link alternative to LinkButton, I'd prefer if we don't duplicate the behavior 🙂
>
> I'd suggest adding a BaseNavigation.AnchorButton component and moving the .match check to Navigation.Button.
>
> We can also omit the check whatsoever and just allow passing href prop to Navigation.Button — if it's passed, we render BaseNavigation.AnchorButton, otherwise we render BaseNavigation.LinkButton.

Ahh, I forgot we had `AnchorButton`! Yes, your approach is much nicer.

Gonna restart on a new PR so I'll close this one

PR #211, opened 2020-02-25
Support external links inside Navigation.Button

joshwcomeau:
Following up from #210, this PR allows us to put external links inside Navigation.Button.

### DIFF
--- a/src/components/skeletons/BaseNavigation/BaseNavigation.js
+++ b/src/components/skeletons/BaseNavigation/BaseNavigation.js
@@ -5,6 +5,7 @@ import { Link } from "gatsby"
 import { MdArrowForward } from "react-icons/md"
 
 import { LinkButton } from "../../LinkButton"
+import { AnchorButton } from "../../AnchorButton"
 import { visuallyHidden } from "../../../utils/helpers"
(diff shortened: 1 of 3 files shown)


joshwcomeau:
@Elanhant looks like the Chromatic test is doing its job! I added a story and it seems to have a problem with it 😅 

I followed the link and I think I need an account... are there docs somewhere that explain how to update Chromatic tests?

src/components/skeletons/BaseNavigation/BaseNavigation.js
```diff
@@ -299,6 +300,27 @@ BaseNavigation.LinkButton = ({
...
+  const { isInverted } = BaseNavigation.useNavigationContext()
+
+  return (
+    <AnchorButton
+      to={linkTo}
+      size={size}
+      css={baseStyles.button(isInverted)}
+      {...rest}
+    >
+      {children} {icon && <MdArrowForward />}
```
Elanhant: I understand that you've just copied this over from `BaseNavigation.LinkButton`, but do you mind changing this to pass `rightIcon` prop to `AnchorButton`? 🙂

Issue #205, opened 2020-02-17
Create Component Checklist


Elanhant:
We should think about having some kind of a checklist to help identify if a component is ready for release. This could help us to achieve some level of consistency as well as increase our confidence in the code we ship.   

First things to come to my mind are:
* The component itself (e.g. `NewComponent`) in its own folder in `src/components`
* A type for component props (`NewComponentProps`)
    * Any "variant" prop should have its own type defined and exported, e.g. 
      ```typescript
      export type NewComponentSize = 'S' | 'M' | 'L'`
      ```
*  An `index.ts` file in the component folder that re-exports the component as well as the props type (and any "variant" prop types) as well as any other functionality deemed useful to be exposed (note that the index file MUST NOT have a default export)
* A story (or stories) for the component (`NewComponent.stories.tsx`)
    *  In most cases should contain a "default" story that showcases the component with the minimum setup
    * If applicable, should have a "sandbox" story which utilises Storybook knobs addon to allow teh user to customize props
    * Any "variant" prop should have its own story to showcase every variant of the prop (for example [see Button stories](https://reverent-keller-9597e8.netlify.com/?path=/story/buttons-button--variants))
    * All stories should be treated as user-facing, meaning we should take a11y in account when writing their code (the most basic mistakes will be caught anyways if we introduce [axe testing](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer#axetest))
*  Unit tests (`src/components/NewComponent/__tests__/NewComponent.test.tsx`)
    * Should use `react-testing-library` if possible
    * ? some components might benefit from snapshot matching
* `src/index.ts` (and, for now, `src/index-ts-only.ts`) should re-export everything from the component's `index.ts` file (`export * from "./components/NewComponent`)

@gatsby-inc/cloud-frontend what are your thoughts on this?

@JeremyTheModernist @fk I have only covered the technical part, but I'm sure there could be something design-related that we could add to this list. At my previous job, we had InVision projects incorporated in each component's story; perhaps we can achieve something like this for Figma (looking at [this addon](https://www.figma.com/resources/extensions-and-apis/storybook-figma-addon/)).

mfrachet:
I think it's a good idea to define a way to create components and to make sure that we all adopt the corresponding style. Adding some a11y tests on these could be really helpful since they are the "building blocks" of .com. (but you have opened an issue / PR with it I think, I'm going to check)

[...]

PR #212, opened 2020-02-25
feat: SidebarNav in TypeScript

Elanhant:
This PR
* rewrites `SidebarNav` in TypeScript
* fixes invalid HTML (we had `<ul>` as a direct child of another `<ul>`, while only `<li>` is supported)
* applies bold font weight to current nav item for better a11y in accordance with [guidelines on using colors](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G182): 
    > To use this technique, an author incorporates a visual cue in addition to color for each place where color alone is used to convey information. Visual cues can take many forms including changes to the font style, the addition of underlines, bold, or italics, or changes to the font size.
* adds some unit tests
* rewrites `SidebarNav` stories to CSF and adds stories suggested in #205 

### Images
`SidebarNav` with changes from this PR (left/former) compared to the current implementation (right/latter); don't pay attention to fonts, the second screenshot was taken from static build which has no access to Futura font:
Screen Shot 2020-02-25 at 9 58 08 AMScreen Shot 2020-02-25 at 9 58 30 AM


src/components/SidebarNav/__tests__/SidebarNav.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    )
+  })
+
+  it(`applies correct aria-current values to links`, () => {
+    ;(global as any).___navigate = jest.fn()
+    const { getByText } = render(Basic())
+
+    fireEvent.click(getByText("Integrations") as HTMLElement)
+
+    expect(getByText("Integrations")).toHaveAttribute("aria-current", "page")
```
mfrachet: Testing against accessibility attributes looks to be a really good practice to me. Thank you for this! 👍

src/components/SidebarNav/__tests__/SidebarNav.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    fireEvent.click(getByText("General") as HTMLElement)
```
mfrachet: WYT of splitting this in another file since you'll make other assertions based on this?

An approach that I really like is AAA (Arrange, Act, Assert) in that order :). That helps me write test in the same way and consistently.

For example, in this situation we are more in a way like "Arrange, act, Assert, Act, Assert" 😄 .

Let me know if you think it's a good idea :) . Here's a link on AAA https://github.com/testdouble/contributing-tests/wiki/Arrange-Act-Assert https://github.com/testdouble/contributing-tests/wiki/Arrange-Act-Assert

src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -0,0 +1,211 @@
...
+    >
+      {to ? (
+        <Link
+          to={to}
+          onClick={onClick}
+          css={(theme: Theme) => ({
+            color: `inherit`,
+            lineHeight: theme.lineHeights.dense,
+            textDecoration: `none`,
+          })}
```
mfrachet: WYT of extracting the css prop here? I agree on the fact that if the css is not too heavy, it's okay to put it in the render output.

However, in that situation, we are in a ternary and the css props is what takes most spaces (it descreases readability for me, but that's maybe a matter of taste? 😅 )

So it's maybe a subjective comment 🤔

src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -0,0 +1,211 @@
...
+          aria-current={current}
+        >
+          {label}
+        </Link>
+      ) : (
+        label
+      )}
+      {children}
```
mfrachet: I'm wondering about the need of `children` when label is provided and is "most" of the SidebarItem thing?

Elanhant: Given that I couldn't find any usages of `SidebarNav` in Cloud that relied on `children` prop, and taking into account tree shaking, I'm going to drop support for `children` and stop exposing `SidebarNav.*` components.

src/components/SidebarNav/SidebarNav.tsx
Elanhant: Yeah good point! I'll update this :)

src/components/SidebarNav/__tests__/SidebarNav.test.tsx
Elanhant: Hm I'm not sure If I understand what do you mean 😅 

Do you think we should move `fireEvent.click(getByText("General") as HTMLElement)` and assertions after this into a separate test? Wouldn't this mean that we are loosing the ability to assert that the previously selected nav item has lost `aria-current` attribute?
```javascript
expect(getByText("Integrations")).not.toHaveAttribute(
  "aria-current",
  "page"
)
```

mfrachet: TL;DR: I think writing two tests in that case could be a good idea:

```typescript

  it(`applies correct aria-current values to "General" content when clicking "General"`, () => {
    ;(global as any).___navigate = jest.fn()
    const { getByText } = render(Basic())

    fireEvent.click(getByText("General") as HTMLElement)

    expect(getByText("Integrations")).not.toHaveAttribute(
      "aria-current",
      "page"
    )
    expect(getByText("General")).toHaveAttribute("aria-current", "page")
    expect(getByText("Site Details")).toHaveAttribute(
      "aria-current",
      "location"
    )
  })

  it(`applies correct aria-current values to the page when clicking "Environment variables"`, () => {
    ;(global as any).___navigate = jest.fn()
    const { getByText } = render(Basic())

    fireEvent.click(getByText("Environment Variables") as HTMLElement)

    expect(getByText("General")).toHaveAttribute("aria-current", "page")
    expect(getByText("Site Details")).not.toHaveAttribute(
      "aria-current",
      "location"
    )
    expect(getByText("Environment Variables")).toHaveAttribute(
      "aria-current",
      "location"
    )
  })
```

As you may see, we are in two different "flow":

- one is against the "General" element selection
- one is against the "Environment variable" element selection

--- 

What I wanted to express before was that it's not a good practice to make the following **in a single test**:

```diff
+ Prepare the tests (this step is good) => "Arrange" step
+ Fire an event (this step is also good) => "Act" step
+ Make an expectation (this step is also good) "Assert" step
- Fire an event (woops, we already have our three AAA steps, so we should not do this)
- Make an expectation (woops, we already have our three AAA steps, so we should not do this)
```

The solution is to break the tests in two to make them smaller and more focused :)

Let's take the following example:

```javascript
// Good
it("should blah blah", () => {
  // Arrange
  const { getByText } = render(Basic())

  // Act
  fireEvent.click(getByText("Environment Variables") as HTMLElement)

  // Assert
  expect(getByText("General")).toHaveAttribute("aria-current", "page")
})

// Not that good because we break the AAA order. The solution is to break it down in 2 tests
it("should blah blah", () => {
  // Arrange
  const { getByText } = render(Basic())

  // Act
  fireEvent.click(getByText("Environment Variables") as HTMLElement)

  // Assert
  expect(getByText("General")).toHaveAttribute("aria-current", "page")

  // Act
  fireEvent.click(getByText("Environment Variables") as HTMLElement)

  // Assert
  expect(getByText("General")).toHaveAttribute("aria-current", "page")
})
```

All of this is debatable, it's just a habit of trying to test a specific unit in this kind of tests (note that this is not true for E2E tests).

Elanhant: Oh I see, thanks for the detailed explanation!

PR #213, opened 2020-02-27
fix: remove static properties from Input form fields

Elanhant:
This PR is the first in a series of changes needed to get rid of static properties from our components to enable Webpack tree shaking. It only covers `InputFieldSkeleton`, `InputField` and `InputFieldBlock` components, as well as makes some updates in `FormFieldSkeleton` and `FormField` which are necessary for this work.

The changes can be generally described as follows:
* `FormFieldSkeleton` now exports its compound parts in addition to specifying them as static properties:
  ```javascript
  export const FormFieldSkeletonLabel = props => {
    /* ... */
  }
  FormFieldSkeleton.Label = FormFieldSkeletonLabel
  ```
  This is done to avoid rewriting ALL of form skeletons and fields; we'll omit static properties once all of skeleton components rely on named exports. 
* `FormField` is changed in a similar fashion to `FormFieldSkeleton`, plus I've added some hooks for styling field errors, labels and hints so that we can use components from respective skeletons instead of generic ones from `FormFieldSkeleton`
* `InputField` now has no static properties; its compound components are exposed as named exports, for example `InputFieldControl` instead of `InputField.Control`
* Because of dropping static properties, it doesn't make much sense to use default exports, so compound components are using named exports instead (similar to ReachUI)

I also have to warn that we these changes won't enable proper tree shaking by themselves, since some of form components rely on `React.forwardRef`, which also cannot be tree-shaken 🤦‍♂ I am working on a babel plugin to fix this issue, it should be ready relatively soon.

### DIFF
--- a/src/components/form-skeletons/stories/storyUtils.ts
+++ b/src/components/form-skeletons/stories/storyUtils.ts
@@ -1,0 +1,20 @@
+export function getGroupFieldStoryOptions() {
+  return [
+    `Assire var Anahid`,
+    `Francesca Findabair`,
+    `Fringilla Vigo`,
+    `Ida Emean aep Sivney`,
+    `Keira Metz`,
+    `Margarita Laux-Antille`,
+    `Philippa Eilhart`,
+    `Sabrina Glevissig`,
+    `Sheala de Tancarville`,
+    `Triss Merigold`,
+    `Yennefer of Vengerberg`,
+  ].map(name => {
+    return {
+      label: name,
+      value: name.toLowerCase().replace(/\s/g, `-`),
+    }
+  })
+}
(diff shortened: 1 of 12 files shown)


src/components/form-skeletons/stories/storyUtils.ts
```diff
@@ -0,0 +1,20 @@
+export function getGroupFieldStoryOptions() {
```
mfrachet: ❓ what is it use for?

Elanhant: This was going to be shared across multiple story files for each field skeleton, but then I decided to reduce the scope of this PR to just input field :)

mfrachet: Thanks 😄

PR #219, opened 2020-03-02
Tooltip component

Elanhant:
This PR adds a `Tooltip` component based on ReachUI's [`@reach/tooltip`](https://reacttraining.com/reach-ui/tooltip/) and the work I've done in #156. 

Changes compared to the previous PR:
* ReachUI's component got rewritten in TypeScript and their type definitions changed drastically, so I had to update my code to reflect that
* No more children functions for `Tooltip` — just render whatever element you want inside the component, as long as it accepts a reference to an underlying DOM node. This mimics the original behaviour of `@reach/tooltip`
* In order to make `Tooltip` tree-shakeable, I had to replace importing default styles from `@reach/tooltip` with applying them locally via Emotion; sadly, `@reach/tooltip` started spamming with warnings that the styles are not loaded, so I had to come up with a hack component to suppress those
* `ariaLabel` prop renamed to `aria-label` for uniformity and to avoid possible confusion
* Stories rewritten in CSF


src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,177 @@
...
+    <button>Hover on me!</button>
+  </Tooltip>
+)
+
+const LONG_TEXT = `Lorem ipsum dolor sit amet consectetur adipisicing elit. Quo officia recusandae nisi magni, dolore laboriosam maiores suscipit perspiciatis. Perspiciatis quod ipsum corporis officia necessitatibus, doloribus fuga culpa. Unde, molestiae repellendus.`
+export const WithLongTooltipText = () => (
+  <Tooltip label={LONG_TEXT}>
+    <button>Hover on me!</button>
+  </Tooltip>
+)
```
mfrachet: I love stories of that shape 👍 !

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * To mark our Tooltip component as safe for tree-shaking, we have to apply global styles
+ * from @reach/tooltip/styles.css locally via Emotion instead of importing them
+ *
+ * However, @reach/tooltip will display a warning if we do so,
+ * and this is what this hack component is for.
+ *
+ * It MUST be a component, so that it can be rendered before/at the same time as the tooltip trigger
+ */
+function DisableReachStyleCheck() {
```
mfrachet: This is tricky 😅 and good to know!

mfrachet: Or it could be a custom hook btw? It doesn't matter I think but I just to make sure I understand every implications :)

src/components/Tooltip/TooltipContent.helpers.ts
```diff
@@ -0,0 +1,66 @@
+import { Position } from "@reach/tooltip"
+import { TooltipPosition, PRect } from "./types"
+import { TOOLTIP_POINTER_SIZE } from "./TooltipPointer"
+
+export const tooltipPosition: Record<TooltipPosition, Position> = {
```
mfrachet: ℹ️ This is a good place for unit testing https://www.notion.so/Testing-scope-and-thoughts-631ba7c24af84780bead0701a92b69ed#19012dbc33f94852bc392b7498a286c3 unit testing section 😄

src/components/Tooltip/Tooltip.tsx
Elanhant: If we were to use it a hook, we'd have to do this inside `Tooltip` component, which would break the "mount" order required to actually suppress the warning. The warning is triggered by `useTooltip` in its own `useEffect`, so we need to be sure that `--reach-tooltip` is set **before** that check runs. Calling another hook at the same level as `useTooltip` does not guarantee this; but rendering a child component does.

If it makes it clearer, I can add this explanation to the comment for this `DisableReachStyleCheck ` component 🙂

src/components/Tooltip/TooltipContent.helpers.ts
Elanhant: I'll add an issue to write some unit tests for this 👍 

Also, since it's an open source repo, I think we should avoid posting links that are not publicly available 😅

PR #223, opened 2020-03-09
Fixing a11y issue on navbar

mfrachet:
aria-expanded was "active" and "undefined".

It's now "true" or "false" on the burger icon :)


Elanhant:
👍

PR #229, opened 2020-03-11
Cleanup form components

Elanhant:
Lots of cleanup in styled form components to make code more DRY and reduce the amount of layers between styled and skeleton form fields.

I'm also trying to figure out the best way to handle stories for form elements, so far my thinking is along these lines:
* Detailed stories for styled "block" fields as they ideally should be the most used part of our forms solution. The stories would include many variations, such as "with hint"/"with error" states which should also help with visual regression testing — more chances to catch unexpected changes to elements that are not usually visible "on mount"
* The bare minimum story coverage for skeleton components — just "basic" and "sandbox" stories to showcase how to compose these components together
* Only basic stories for "Connected" fields — they for the most part rely on the styled "block" fields, so the only thing we need to showcase is how to use them in a form powered by Formik.


mfrachet:
This PR is a bit hairy, I will need some time to review it carefully 😅

Elanhant:
> Everything looks working on the storybook. I didn't see anything that is blocking me from merging. I think we should merge this quickly since it starts getting huge 😅 and iterate when needed. WYT?

Oh I have no intention of bloating this any more than this 😄 Any story updates will definitely be made in separate PRs.

PR #230, opened 2020-03-11
feat: update gatsby-design-tokens

Elanhant:
Updates `gatsby-design-tokens` to v2 and includes relevant changes to our theme configuration (see [changelog](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-design-tokens/CHANGELOG.md#200-2020-02-10)).

Next steps to be covered in separate stories:
1. Try to use default theme exposed by `gatsby-design-tokens`
1. Use Inter font


src/theme/transition.ts
```diff
@@ -1,46 +1,4 @@
...
-// we change slow value to `500ms` and add some additional options
-const transition: Transitions = {
-  ...baseTransition,
-  curve: {
-    ...baseTransition.curve,
-    fastOutLinearIn: `cubic-bezier(0.4, 0, 1, 1)`,
-  },
-  speed: {
-    ...baseTransition.speed,
-    snail: `1000ms`,
```
fk: Found that we use `snail` in an example in the README. Since I'm checking this out locally, will try if I can push a fix for that.

fk: Yep that worked :) https://github.com/gatsby-inc/gatsby-interface/pull/230/commits/0a134986080fbb432ac2030a2dfd5e9fc1606638

fk: Searched more, found `docs/design-tokens.md`, bumped that, too https://github.com/gatsby-inc/gatsby-interface/pull/230/commits/acd7f372621574739eed236ef903ef82c6d4312c

Elanhant: Thank you @fk 💜

fk: 🙌 💜

fk:
I think I got all the changes for the theme README and `docs/design-tokens.md` ✌️— sorry for the noisy commits!

fk:
@greglobinski I stumbled upon what I think is a typo in https://github.com/gatsby-inc/gatsby-interface/blob/master/docs/design-tokens.md#fontsizes — "phares" — would have fixed but wasn't sure what it should be:

> If you have to use font-size value smaller than 0.75rem (fontSizes[0]) you can use inline values fontSize: "0.5rem", but that's only applicable to text as an graphic element (e.g. badge with number), you should never use a **phares** or body text smaller than fontSizes[0].

PR #231, opened 2020-03-12
Fixing modal leaving when clicking on the overlay

mfrachet:
This pr aims to fix the problem of the overlay not always closing the modals.

To test it, run the storybook and make sure that you clicking anywhere on the overlay logs "Dismissed!" in the console


src/components/Modal/Modal.tsx
```diff
@@ -51,26 +50,22 @@ const Overlay = styled(DialogOverlay)<Props>`
...
-)
+}) => {
+  return (
+    <Overlay
+      initialFocusRef={initialFocusRef}
+      isOpen={isOpen}
+      onDismiss={onDismiss}
+      animation={getBackgroundAnimation(type)}
+    >
+      {React.cloneElement(children as any, props)}
```
mfrachet: ℹ️ Not perfect, but it allows to keep the actual API without having to modify all the codebases

PR #237, opened 2020-03-12
feat: refactor and rewrite Card to TypeScript

Elanhant:
Rewrites `Card`, `CardHeader`, `CardTitle` and `CardContent` (formerly `CardText`) to TypeScript, restricts their respective sets of possible HTML elements and adds a "standalone" state which should allow to use a card inside another card if needed (similar to how build cards work in Cloud).

Some musings from me:
Even though this particular `Card` is currently used in just one place in Cloud, we could use it more extensively. The biggest obstacle I can see is the lack of definition for a "card" component — should it be a very general definition (like it is right now), in which case we should perhaps rely on style primitives and not components? Then we can build more complex and specific cards, like a build card, a settings card etc.  If so, what should be part of the "card" primitive — should we include paddings? They seem to differ a lot across our current card-like components.


fk:
> Then we can build more complex and specific cards, like a build card, a settings card etc. If so, what should be part of the "card" primitive — should we include paddings?

Very good question … both horizontal and vertical padding I assume? 😬 

> They seem to differ a lot across our current card-like components.

^ 💯, and 💯% on me. I'm trying to fix that, and stick to multiples of `8`—instead of 20, 24, 40, 48, probably more? (I didn't check, but IIRC…smh at myself).

fk:
For whatever reason I started wondering if `CardContent` should be `CardBody`…naming things, always fun! 😉 

Anyway—that made me realize that in a project some years ago I chose to include padding for the card header and footer, but decided against including it for the content/body of the card—simply because I found myself using negative margins there _a lot_. Just thinking out loud for now, please bear with me…

src/components/Notification/Notification.helpers.ts
```diff
@@ -25,14 +24,12 @@ type GetVariantStylesFn = (tone: NotificationTone) => ThemeCss
 const variants: Record<NotificationVariant, GetVariantStylesFn> = {
   PRIMARY: tone => {
     return theme => [
-      cardFrameCss(theme),
       {
         background: theme.colors.white,
         borderLeft: `10px solid ${theme.tones[tone].dark}`,
```
fk:
```suggestion
        borderLeft: `${theme.space[2]} solid ${theme.tones[tone].dark}`,
```

This dials down the left border to `4px` width, and should go alongside changing `paddingLeft` to `20px`—will leave another suggestion for that.

src/components/Notification/Notification.helpers.ts
```diff
@@ -25,14 +24,12 @@ type GetVariantStylesFn = (tone: NotificationTone) => ThemeCss
...
-        padding: `${theme.space[5]} ${theme.space[7]} ${theme.space[5]} ${theme.space[5]}`,
-        [theme.mediaQueries.desktop]: {
-          padding: `${theme.space[5]} ${theme.space[7]} ${theme.space[5]} ${theme.space[5]}`,
-        },
+        paddingTop: theme.space[5],
+        paddingLeft: theme.space[5],
```
fk:
```suggestion
        paddingLeft: theme.space[6],
```

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const baseCss: ThemeCss = theme => ({
+  background: theme.colors.white,
+})
+
+const standaloneCss: ThemeCss = theme => ({
+  paddingTop: theme.space[6],
+  paddingBottom: theme.space[5],
+  paddingLeft: theme.space[7],
+  paddingRight: theme.space[3],
```
fk:
```suggestion
  paddingRight: theme.space[7],
```

I might be missing something here! 🤔

src/components/Card/CardContent.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const baseCss: ThemeCss = theme => ({
+  color: theme.colors.grey[50],
+  fontFamily: theme.fonts.body,
+  fontSize: theme.fontSizes[1],
+  lineHeight: theme.lineHeights.body,
+  marginTop: theme.space[5],
+  marginRight: theme.space[5],
+  marginBottom: theme.space[5],
+  marginLeft: 0,
```
fk: @Elanhant re:your musings about a base card component and including padding—seeing this, I think it _might_ come in handy at some point to be able to add a background color to the whole card body…errr, content ;-). Same applies to the header, and also eventually a footer component.

src/components/Notification/Notification.helpers.ts
fk: On second thought, these changes should probably go into a follow-up PR.
While reviewing this one, I noticed a couple more things irt the "Notifications" component specifically that I'd like to address, happy to take care of that once this is merged:

- `borderLeft` and `paddingLeft` (the ones I left suggestions for in here)
- nudge variant colors a bit (making the `BRAND` ones a bit friendlier and calmer)
- adjust icon sizes (the size that the icons are rendered in is 👍, but all other icons we are currently using contain inner padding, while our custom ones do not)

image

Elanhant: Cool, I'll add an issue for this 👍

src/components/Card/Card.tsx
Elanhant: I actually got these styles from Build cards, which again demonstrates that we're are not very consistent 😆 

Perhaps we should have card variants which would define what paddings to use? I don't expect that we are going to always be able to maintain the same padding, but having a variant should at least restrict us a little bit :)

Elanhant:
> Anyway—that made me realize that in a project some years ago I chose to include padding for the card header and footer, but decided against including it for the content/body of the card—simply because I found myself using negative margins there _a lot_. Just thinking out loud for now, please bear with me…

Yeah this actually makes a lot of sense, and I think I used to do something similar. It becomes especially obvious when you want to have some kind of a "separator" line in your content which looks better if it stretches from one side of the card to another.

I guess a possible alternative to specifying padding directly in the component is to expose a function that would return "standard" padding values for a card which then could be used to apply them to whatever element you want. I am actually already doing something like this for Build cards in Cloud:
```javascript
// These are some default styles for a card 
// Note that they also include default padding
export const getCardBaseStyles = theme => [
  {
    background: theme.colors.white,
    borderRadius: theme.radii[2],
    boxShadow: theme.shadows.raised,
    transition: `box-shadow ${theme.transitions.speed.default} ${theme.transitions.curve.default}`,
    "&:hover, &:focus, &:focus-within": {
      boxShadow: theme.shadows.floating,
    },
  },
  // Apply padding by default
  getCardPadding(theme),
]

// And this can be used by card header, content, footer or anything else really
// to adjust their paddings
export const getCardPadding = theme => ({
  paddingTop: theme.space[6],
  paddingBottom: theme.space[5],
  paddingLeft: theme.space[7],
  paddingRight: theme.space[3],
})
```

We still are going to have the problem of inconsistent values with our current cards, but we could use my suggestion of variants here as well. What do you think @fk?

src/components/Notification/Notification.helpers.ts
Elanhant: An issue for your suggested changes: #239

Elanhant:
It also seems like we already have something similar to what I suggest: https://github.com/gatsby-inc/gatsby-interface/blob/master/src/theme/styles/card.ts

src/components/Notification/Notification.helpers.ts
fk: Thank you! Resolving this (no need to fix)!

src/components/Card/Card.tsx
fk: Yep,  defs 👍 for variants. Will try to boil things down!
… and again, resolving!

src/components/Notification/Notification.helpers.ts
fk: Will add a follow up PR for these ✌️

fk:
> I guess a possible alternative to specifying padding directly in the component is to expose a function that would return "standard" padding values for a card which then could be used to apply them to whatever element you want. 

Sounds good…

> It also seems like we already have something similar to what I suggest:

…and looks good; I'll dig into that for #239.

Elanhant:
I am going to wait till #247 is merged before merging this PR — I think it will be easier to resolve any conflicts (especially when it comes to Chromatic)

fk:
💯 👍

Elanhant:
@fk I've decided to remove `Card` component completely in favour of having `cardStyles` in theme ([usage example here](https://github.com/gatsby-inc/gatsby-interface/pull/237/commits/18d96b0bdfb27dc182a92e6bd8cdbe991d4f0a17#diff-6dfcb101728a40dc8db10f36bef90009R77), and the definition can be found [here](https://github.com/gatsby-inc/gatsby-interface/pull/237/commits/18d96b0bdfb27dc182a92e6bd8cdbe991d4f0a17#diff-42c090954509eda0d00c8ebb63467ec1R208)).

mfrachet:
I only see one risk with an approach opened and based on style: we can tweak it extensively and break potential cohesion. An approach that I really like and that you have mentioned is to find a definition for a card component: what is a Card what is its aim and how should we display the content? 

Just like we did for the Button component: we have some variants, we have tones that prevents us of having too many different kind of buttons

I'm wondering about studying the material design spec about cards. Without pushing everything too far (do we really need this? 🤔).

I'm thinking about cohesion in our designs (UI). I love the idea of extending things and tweaking stuff. And I'm also wondering what are your thoughts on this?

Elanhant:
I feel like currently we use the idea of "card" not as a component but as a semantic style — the only thing that unites `BuildCard` in Cloud and `Notification` and `SettingsCard` in gatsby-interface is their border and background styles (even paddings aren't consistent!). This is why I decided to omit the `Card` component completely in favour of a style object. 

> I only see one risk with an approach opened and based on style: we can tweak it extensively and break potential cohesion

This is already happening with the `Card` (and `Card`-like) components 😅We have lots of places where we outright override default paddings. 

We do need to figure out how do we use (and want to use) cards, but this should probably happen in a different place, not in this PR discussion :) 

I'll create an issue where we can properly discuss what a Card should look like.

mfrachet:
That sounds good to me for now :) . Let's make moves and deal with it later. Maybe @fk will have potential new information with the design system approach? (haha! Throwing the 🥔 to yah 😝)

fk:
🥔 🤗 😅 

Left a comment in #251—let's audit! We can, need, should def. consolidate.

One quick thought re cohesion—I think #251 will get us close to two different card paddings … ~`default` and `large`. I also suspect we need something along the lines of https://material.io/design/layout/applying-density.html.

PR #244, opened 2020-03-18
feat: Heading rewritten to TypeScript

Elanhant:
Rewrites `Heading` (and `BaseHeading`) to TypeScript.


src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,60 @@
...
+  },
+  variant === `EMPHASIZED` && {
+    fontWeight: theme.fontWeights.extraBold,
+  },
+  variant === `LIGHT` && {
+    fontWeight: 100,
+    textTransform: `uppercase`,
+    color: theme.tones[tone].dark,
+  },
+]
```
fk: Seeing this reminded me of a question again:

I understand that using https://theme-ui.com/theme-spec#variants with the `variant` prop would require us to opt in to the `sx` prop…_that aside_, I was wondering if we should define these variant styles in the theme, still?

fk: LOL while I was writing it dawned on me that the `variant` styles use the `sx` ✨, too. 🤐

Elanhant: Well we can indeed list the variant styles in our theme (just without using the `sx` prop), but I have some concerns regarding this:
* How do we decide which components should have their variants defined in themes? Should it be just the most basic stuff like buttons, text and headings? What about Notifications or Tabs? 
* Putting component-specific styles into theme is going to affect tree shaking — IIRC it is not a trivial task to remove unused object properties, and a theme is in the end an object. It shouldn't be a big problem if we stick to storing variants just for primitives though.

Elanhant: You know what? I feel like card padding would be a great candidate to be stored in the theme object, as it is quite abstract 😄

fk: 💯 , fully agree. (Also sorry for not replying earlier. I even started writing a response … it's still there, in one of those _other tabs_ :D)

PR #245, opened 2020-03-18
fix: do not import Reach dialog css

Elanhant:


Elanhant:
> Oh wow, that's the fix for the modal woes?

Yes, if we are talking about the same woes :) 
Looks like my efforts at ensuring proper tree shaking resulted in CSS import statements not being properly processed by Cloud 😓 

But we already switched from direct import to using custom CSS for `Tooltip`, so this should work for modals too.

fk:
👍 👍

PR #247, opened 2020-03-19
Setup MDX docs support

Elanhant:
We've already been using [Storybook Docs](https://github.com/storybookjs/storybook/blob/master/addons/docs/README.md#storybook-docs) to display our theme scales ([font sizes, for example](https://reverent-keller-9597e8.netlify.com/?path=/docs/theme--fonts-sizes)).

This PR adds support for MDX docs and a couple of basic examples (`Intro.stories.mdx` and `Usage.stories.mdx`). With these we can write more elaborate documentation for our components. The only problem so far is that providing custom components for Markdown elements (e.g. headings) doesn't seem to work at the moment.

I have also upgraded all of Storybook dependencies to v6 (which is currently in alpha, but contains a lot of fixes unavailable in v5.3); because of that I also had to remove `storybook-readme` which seems to be incompatible with v6 and which would become obsolete if we start using anyway.

I have also played a little bit with theming Storybook — can't say there are a lot of options there, but perhaps @fk would be interested in it :) You can see it in action [here](https://5d4c6885a869c600201d3c8b-pbmgxjooxr.chromatic.com/?path=/story/buttons-anchorbutton--variants).


.storybook/manager.js
```diff
@@ -0,0 +1,26 @@
...
+  sidebarAnimations: true,
+  enableShortcuts: true,
+  isToolshown: true,
+  selectedPanel: undefined,
+  theme: create({
+    ...themes.normal,
+    base: `light`,
+    fontBase: theme.fonts.body,
+    fontCode: theme.fonts.monospace,
+    colorSecondary: theme.colors.gatsby,
```
fk: ✨

.storybook/preview.js
```diff
@@ -98,23 +86,46 @@ if (process.env.NODE_ENV === `test`) {
...
-addParameters({
-  options: {
-    addonPanelInRight: true,
+const viewports = {
+  mobile360x640: {
+    name: "Mobile 360 x 640",
+    styles: {
+      width: "360px",
+      height: "640px",
+    },
```
fk: Hello there! 😻 (full n00b mode activated).
I guess I will add a bunch of those … wondering if there are presets available for the usual suspects.

Elanhant: Well the viewports add-on has [some default viewports](https://github.com/storybookjs/storybook/blob/master/addons/viewport/src/defaults.ts), we can easily extend them instead of replacing :)

__stories__/about/Intro.stories.mdx
```diff
@@ -0,0 +1,8 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+import { Heading, Text } from '../../src'
+
+<Meta title="About Gatsby Interface/Intro" />
+
+# Intro
+
+Gatsby Interface is a component library initially built for Gatsby Cloud
```
fk: IMO we really need to add some color here ;)

__stories__/about/Usage.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+1. Initiate a new Gatsby Site
+    ```bash
+    gatsby new <my-site-name>
+    ```
+1. Install Gatsby Interface and its peer dependencies:
+    ```bash
+    cd <my-site-name>
+    ```
+    <code className="language-bash">
+    yarn add {peerDepsList}
```
fk: 😻

.storybook/preview.js
fk: 👍

Elanhant: Added [INITIAL_VIEWPORTS](https://github.com/storybookjs/storybook/blob/master/addons/viewport/src/defaults.ts#L3) to our list.

fk: Naaaaisssss! That is an extensive list :)

.storybook/preview.js
```diff
@@ -1,12 +1,12 @@
...
-import { configure, addDecorator, addParameters } from "@storybook/react"
-import { addReadme } from "storybook-readme"
+import { addDecorator } from "@storybook/react"
 import { withKnobs } from "@storybook/addon-knobs"
 import { withConsole } from "@storybook/addon-console"
 import { withA11y } from "@storybook/addon-a11y"
 import { action } from "@storybook/addon-actions"
-import withTheme from "./withTheme"
+import { INITIAL_VIEWPORTS } from "@storybook/addon-viewport"
+import { withTheme, docsMDXComponents } from "./theming"
```
fk:
```suggestion
import { withTheme, docsMDXComponents, storybookThemeLight } from "./theming"
```

.storybook/preview.js
```diff
@@ -98,23 +86,30 @@ if (process.env.NODE_ENV === `test`) {
...
+  ...INITIAL_VIEWPORTS,
+}
 
-//const req = require.context("../__stories__", true, /\.stories\.js$/)
+export const parameters = {
+  viewport: { viewports: viewports },
+  docs: {
+    // For some reason currently these overrides do not work
+    // Seems that we are not alone in this: https://github.com/storybookjs/storybook/issues/9968
+    components: docsMDXComponents,
```
fk:
```suggestion
    components: docsMDXComponents,
    theme: storybookThemeLight,
```

PR #249, opened 2020-03-24
chore: fix dependencies to support Storybook Docs theming

Elanhant:
With this PR we should get Storybook Docs use our own components for text and headings.

A couple of notes:
*  `DocsCodeOrSourceMdx` is introduced because Storybook for some reason does not apply `fontCode` style to `<code>` tag in MDX 🤷‍♂ 
* Custom components for MDX are rendered with our `ThemeProvider`, otherwise they would receive their theme scales from Storybook


fk:
Yay! 🙏

PR #250, opened 2020-03-24
feat: Modal style fixes and CSF stories

Elanhant:
Most of this PR is just rewriting Modal stories in a CSF format plus:
* Removing padding rules from `ModalCard` and `ModalPanel` — seems like in all but one usages of `ModalCard` in Cloud we're resetting padding rules by passing `css={{ padding: 0 }}`; this kinda hints that we were wrong to include default padding 😅 
* Use object styles instead of template literals in modal-related components
* A tiny bit of types cleanup, I've also exposed all of the prop types via index file

@mfrachet I have a feeling you might be interested in these changes 🙂.

P.S. The next step after this would be to start moving some of the modal presets and components from Cloud to this repo, I'll add an issue for this later.


Elanhant:
> It looks that you're way better than me to deal with this Modal stuff! They've got me in trouble for so long 😓

Don't be so hard on yourself 😄 All I did was combine a couple components into one :)

PR #254, opened 2020-03-25
Script to scaffold components

Elanhant:
Adds a small script for scaffolding a new component, inspired by the idea of a Component Checklist (#205); this script does the following:
* Generates a file for the component itself, `<component-name>.tsx`, with stubs for props type, styles and even some variants
* Generates a story file, `<component-name>.stories.tsx`, which follows the suggestions from #205
* Generates an index file, `index.ts`, which reexports everything from the component file
* Adds export statements to library index files: `./src/index.ts` and `./index-ts-only.ts`.

The script can be used like this:
```bash
yarn scaffold:component TestComponent
```

Everything that `scaffold:component` generates should be working out of the box and be ready to shipped (though please avoid shipping the stubs 😅)

An example for a component generated by this script [can be found here](https://github.com/gatsby-inc/gatsby-interface/commit/1f89996b09cbf5d5267c5a39959adc4317bc7dfb).


fk:
^ Was thinking the same as Marvin—how about adding the (relevant parts of the) PR description to the README?

Elanhant:
@mfrachet @fk readme updated :)

PR #256, opened 2020-03-26
Add "concealed value" component

amberleyromo:
# Purpose

Add a "concealed value" component to gatsby-interface, e.g. for environment variables.

Went off a similar pattern in 1pass.

# Gif

concealed-value-component


amberleyromo:
Chatted with @marcysutton and also got some good feedback from the a11y Slack channel:

- I had followed the 1Password pattern, but per the a11y Slack channel (and I agree), the action dropdown seems overmuch. If we do find use to expand the number of actions, it may be a useful pattern later. For now, I've simplified to a Copy button and a Conceal/Reveal button. So I pulled out all the `@reach/menu-button` stuff, etc
- @marcysutton while from my research it seems like there's nothing in spec to forbid usage of fieldset/legend outside of a form, it looks like it has longstanding bugs with using flexbox/grid inside of it. So overall while it's not **forbidden**, there are reasons it doesn't seem ideal either.
- After chatting with Marcy we did agree that announcing changes using aria live region would probably make sense and is worth exploring. Marcy brought up it may be problematic to create different instances of it across many usages, so we should explore creating a root announcer component. But, that it was worth looking into in future -- created a new issue for it here: #259

## Updated

concealed-value-component-demo-2

PR #257, opened 2020-03-27
Allow `Navigation` to link to external URLs

joshwcomeau:
In the Will It Build project, we have 1 link (API Playground) that needs to use an `<a>` tag rather than a `<Link>`, since the page is outside of the Gatsby app (links to `willit.build/api-playground`, but we're using Netlify rewrites to actually show a Graphiql playground)

I tried to follow the pattern established with @Elanhant for the `NavigationButton`; `BaseNavigation` has two separate primitives, but `Navigation` looks at the `linkTo` to determine which should be used.

### DIFF
--- a/src/components/BaseNavigation/BaseNavigation.tsx
+++ b/src/components/BaseNavigation/BaseNavigation.tsx
@@ -6,6 +6,7 @@ import { MdArrowForward } from "react-icons/md"
 
 import { LinkButton, LinkButtonProps } from "../LinkButton"
 import { AnchorButton, AnchorButtonProps } from "../AnchorButton"
+import { BaseAnchor } from "../BaseAnchor"
 import useOnClickOutside from "../../utils/hooks/useOnClickOutside"
 import baseStyles from "./BaseNavigation.styles"
 import { visuallyHiddenCss } from "../../stylesheets/a11y"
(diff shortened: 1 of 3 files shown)


src/components/BaseNavigation/BaseNavigation.tsx
```diff
@@ -294,14 +298,37 @@ export type BaseNavigationItemLinkProps = Omit<GatsbyLinkProps<any>, "ref"> & {
   item: BaseNavigationItem
 }
 
+export type BaseNavigationItemAnchorProps = Omit<
+  React.HTMLProps<HTMLAnchorElement>,
```
Elanhant: You can use `BaseAnchorProps` instead of `React.HTMLProps<HTMLAnchorElement>` here :)

As a side note, I can't wait for React to stop forwarding refs by default so that we don't need these omits 😃

src/components/Navigation/Navigation.js
```diff
@@ -91,14 +91,23 @@ Navigation.Item = delegated => {
...
     mobileNavMediaQuery,
   } = BaseNavigation.useNavigationContext()
 
+  const isExternal = getIsExternalLink(item.linkTo)
+
+  const Component = isExternal
+    ? BaseNavigation.ItemAnchor
+    : BaseNavigation.ItemLink
+
+  console.log(Object.keys(BaseNavigation), Component)
```
Elanhant: Seems like a leftover comment

joshwcomeau: Good catch!

joshwcomeau:
Thanks for the lightning-speed review!

PR #258, opened 2020-03-27
feat: Tabs rewritten to StepIndicator

Elanhant:
@fk and I have been discussing how our current usage of `Tabs` in Cloud onboarding flow should really be a step indicator (like in a form wizard). Based on those discussions, I have rewritten `Tabs` in gatsby-interface to `StepIndicator` and removed the original component, as it is only ever used in the onboarding flow now.

I have also cleaned it up a bit by removing unused variants and replacing `TabFigure` with [CSS counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters).

Next steps once this is shipped would be to make the new component a proper step indicator, which entails changes in how it looks and behaves, and improvements in a11y.


Elanhant:
> * all the `StepIndicatorStep` can be "active" at the same time, from my perspectives, this could break consistency
> * the previous point could be error prone since it's up to the developer to manage the step of the different StepIndicator
> 
> I'm wondering if we couldn't use an approach like @reach/router or @reach/tabs, relying on the `React.cloneElement` to improve this consistency 🤔

If I understand it correctly, you're suggesting that instead of passing `status` to each tab we should let the `StepIndicator` component accept this prop and take care of applying the correct status to its children via `.cloneElement`?

mfrachet:
Yes, it could be a way to keep consistent the "only one active" at a time :) . If you'd like to experiment with these solution later, we can first merge the step indicator in this situation. It's really well built and works really really gooooood 🚀

Elanhant:
> Yes, it could be a way to keep consistent the "only one active" at a time :) . If you'd like to experiment with these solution later, we can first merge the step indicator in this situation. It's really well built and works really really gooooood 🚀

Thanks! I'll add an issue to look into delegating "active" state to the `StepIndicator` component. 

I feel like it is less trivial to do for step indicator-like components because they are more complex than tabs — tabs usually don't entail many restrictions, while steps indicator usually disables the "incomplete" steps, and might disable previous steps as well. The current implementation allows for such cases since it is very low-level.

PR #266, opened 2020-04-02
feat: StyledModal component

Elanhant:
This is the first iteration in moving and refactoring styled modal components from Cloud.

This PR takes care of the main elements of a modal: header and body.

You can take a look at those in [Chromatic preview](https://www.chromaticqa.com/component?appId=5d4c6885a869c600201d3c8b&name=Modal%2FStyledModal&buildNumber=266&specName=Basic).


fk:
🎉 🎉 🎉

Elanhant:
> * Will we have to use `<StyledModalHeader />` (and friends) or do we plan to expose these as `<Modalheader />` and friends?

My idea is that for now we're going to use `Styled*` notation, since we already have components like `Modal`, `ModalCard` and `ModalPanel` which take care of behaviour and not styling. I hope that when we start migrating to `Styled*` components, it will be more clear how should we name things. After that, the most effort we'll have to make is to find&replace some strings. 🙂 

> * Will we have a `<ModalFooter />` or `<ModalActions />` coming? 😄 . Can't wait for these layout to see the daaaaaayyyy 🌃 🌅 🌆 🌞

Yup, that will be the next step after we start using the components from this PR in Cloud!

PR #267, opened 2020-04-03
chore: exclude Sandbox stories from Chromatic

Elanhant:
We don't really need to perform visual regression testing for "Sandbox" stories since those usually look the same as their "Basic" counterparts. 

This PR excludes all of the existing "Sandbox" stories from Chromatic checks and updates the `.stories.tsx` template accordingly.


Elanhant:
> Thanks! This should probably reduce the chromatic exec time right? This is great 😊

Given that I want to try out viewport snapshots as well, it's not that big of a win 😄 But yeah, the fewer unnecessary tests the better!

PR #268, opened 2020-04-03
feat: DensityProvider

Elanhant:
This is a proposal for handling density inspired by [Material Design](https://material.io/design/layout/applying-density.html#components). Relying on React context instead of passing props allows to control density for multiple components at the same time, satisfying this rule:
> Component density should not be adjusted in isolation

So we should be able to do something along these lines:
```jsx
<DensityProvider density="DENSE">
  <Input type="email" placeholder="Type your e-mail" />
  <Button>Submit</Button>
</DensityProvider>
```
Of course, we'll need to add density support for our existing components :) 

The density scale is very basic, open to suggestion on what it should actually be 🙂 It currently supports three values, `DENSE`, `DEFAULT` and `LOOSE`:

Screen Shot 2020-04-02 at 8 56 51 PM


Elanhant:
> 
> Are only the forms component subject to density?

Not at all! If you look a the "Usage" section in [Material Design docs](https://material.io/design/layout/applying-density.html#usage) or [Density components docs](https://material.io/develop/web/components/density/), you'll see that it can also be used for data tables, lists and tabs!

fk:
🎉 🎉 🎉 

> Of course, we'll need to add density support for our existing components :)

Let's do it! 💪

PR #271, opened 2020-04-05
feat: add <Notification variant="SOLID">

fk:
image

https://deploy-preview-271--gatsby-interface.netlify.app/?path=/story/notification--sandbox


src/components/Notification/Notification.tsx
```diff
@@ -141,18 +162,35 @@ function NotificationContent({
...
         minHeight: `auto`,
-        color: theme.colors.grey[40],
+        color:
+          variant === "SOLID"
+            ? theme.colors.whiteFade[60]
+            : theme.colors.grey[40],
         width: theme.space[5],
         marginLeft: theme.space[5],
         fontSize: theme.fontSizes[4],
+        ...(variant === "SOLID"
```
Elanhant, line 186: You can use arrays to add styles based on some condition:
```jsx
<Button
  css={theme => [
    {
      padding: 0,
      // ...
    },
    variant === "SOLID" && {
      ":hover": {
        background: "transparent",
      },
    },
  ]}
/>
```

src/components/Notification/Notification.tsx
```diff
@@ -141,18 +162,35 @@ function NotificationContent({
   )
 }
 
-function NotificationDismissButton({ label = `Close` }: { label?: string }) {
+function NotificationDismissButton({
+  label = `Close`,
+  variant,
```
Elanhant, line 167: We can put `variant` into `NotificationContext` like we do with `onDismiss`, then component users won't have to pass it twice if they ever need to use `NotificationDismissButton`.

src/components/Notification/Notification.tsx
```diff
@@ -73,20 +73,29 @@ export default function Notification({
     <NotificationContext.Provider value={{ onDismiss: onDismissButtonClick }}>
       <div
         css={(theme: Theme) => [
+          { borderRadius: theme.radii[2] },
```
Elanhant, line 76: This will affect `SECONDARY` variant, won't it (that variant doesn't have rounded corners atm)?

Elanhant: Ah I see
> add theme.borderRadius[2] for all notification variants

In which case we should move this style to `baseCss` declaration :)

fk: 👍 (still catching up :/)

src/components/Notification/Notification.tsx
fk: 💯 D'OH 🤦

fk: ✌️ 🙏

src/components/Notification/Notification.tsx
fk: Hopefully in https://github.com/gatsby-inc/gatsby-interface/pull/271/commits/7d33c522288c37b94a50e811a241c783ff9689ac

src/components/Notification/Notification.tsx
fk: Decided to revert this (695a37d), and not alter the `SECONDARY` variant in any way … like smart people would have done in the first place :-P ;-)

src/components/Notification/Notification.tsx
```diff
@@ -70,35 +74,72 @@ export default function Notification({
...
-    <NotificationContext.Provider value={{ onDismiss: onDismissButtonClick }}>
+    <NotificationContext.Provider
+      value={{ onDismiss: onDismissButtonClick, variant, tone }}
+    >
       <div
         css={(theme: Theme) => [
           variant === `PRIMARY` && theme.cardStyles.frame,
           baseCss(theme),
           getNotificationVariantStyles(variant, tone)(theme),
+          variant === `SOLID` && { borderRadius: theme.radii[2] },
```
Elanhant, line 85: I would put all of your additions here inside `getNotificationVariantStyles`, since the name (and the `tone` prop being passed) kinda suggests that it should take care of all variant-related styles 🙂

fk: That makes sense. Lemme try 👶 👣…

fk: (Ah _soooo much_ 🤦) (I don't mean to be ignorant, I swear! ✌️)

fk: c2f967f

PR #272, opened 2020-04-05
chore: add knob to control `variant` to the <Notification> "Tones" story 

fk:
Extracted from #271

### DIFF
--- a/src/components/Notification/Notification.stories.tsx
+++ b/src/components/Notification/Notification.stories.tsx
@@ -93,6 +93,7 @@ export const Tones = () =>
       key={tone}
       tone={tone}
       content={`Notification tone "${tone}"`}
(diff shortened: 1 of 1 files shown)


Elanhant:
Hm, my idea was that in 99% cases knobs would only be used in "Sandbox" stories; is there a specific reason for adding `variant` to `Tones`?

fk:
Yes, to see all of them at a glance, instead of using the sandbox to click through one-by-one. We could adjust the sandbox, or even use s/th like https://github.com/evgenykochetkov/react-storybook-addon-props-combinations instead?!

Elanhant:
@fk could you please check #279? I hope it can be a good alternative.

fk:
#279 def. is nice! 🎉

Still wondering about visual regression tests :-P ;-) — but #279 def. does scratch my itch, closing this!

PR #274, opened 2020-04-06
feat: realign `<SidebarNav>`

fk:
image

https://deploy-preview-274--gatsby-interface.netlify.com/?path=/story/sidebarnav--basic

- drop `Futura PT` :tada:, adjust `fontSize` to `14px`
- always display SidebarNav icons, not only when the item is `active`
  - make `colors.grey[40]` the default icon color
  - display icons at `20px` via `fontSizes[4]`
- redo existing icons
  - make all icons "listen" to `fontSize` (as opposed to hard coding `height` and `width`)
  - nudge all icons to fit the Material Design icons grid 🏁 
  - new icons for `General`, `Integrations`, `Preview` ✨ 
    - BREAKING CHANGE: mv icon `Preview` -> `Builds` 💥 
  - instead of individual colors, use `currentColor` for `fill` and `stroke`
- add `Reports` icon
- `semiBold` instead of `bold` for `active` `SidebarNav` items; use theme tokens
- link Figma design
- expand `SidebarNav` story example items


Elanhant:
@fk do you mind trying `yarn scaffold:icon` for the new icon assets? Let me know if it does not satisfy some icon requirements; I'd prefer if we had only one source of our icons with shared API provided by `IconSkeleton`.

fk:
Oh … oh oh oh. I. forgot about the template again, sorry Aleksei!
Of course—will update the PR ASAP! ✌️

fk:
Just moved the adjusted icons to the template, and threw in some overrides to make them work with the skeleton defaults.

Reviewing the changes in Chromatic just made me realize that setting `height` and `width` does make them respond to the parent's `font-size` also breaks the `size` prop … d'oh 🤦‍♂.  
I think it's OK to get those in and think about how we can support both? Shouldn't be too hard? 😬   
The main reason for keeping the `1em` stuff around is that we currently change the root font size for large(eeeer) screens—IMO altogether it does currently make sense to do so, given that these icons are just used in the context of `SidebarNav` for now.

PR #275, opened 2020-04-06
Adding Dropdown

mfrachet:
This PR aims to add the `Dropdown` component with the ability to search inside (optional).

_Current caveat: when you fill the full name of an item in the search bar, you can't press Enter anymore. Except if you space, backspace, and press Enter... I don't know why 🤷‍♂ _

## How it works?

It relies on https://reacttraining.com/reach-ui/menu-button/ for the accessibility part and is navigable with a keyboard.

## API

```jsx
export const Component = () => {
  const [selected, setSelected] = useState()

  return (
    <Dropdown>
      <DropdownLabel placeholder="Placeholder for the dropdown">
        {selected}
      </DropdownLabel>
      <DropdownItems>
        <DropdownItem
          onSelect={() => setSelected("First")}
          selected={selected === "First"}
        >
          First
        </DropdownItem>
        <DropdownItem
          onSelect={() => setSelected("Second")}
          selected={selected === "Second"}
        >
          Second
        </DropdownItem>
        <DropdownItem
          onSelect={() => setSelected("Third")}
          selected={selected === "Third"}
        >
          Third
        </DropdownItem>
        <DropdownItem
          onSelect={() => setSelected("Fourth")}
          selected={selected === "Fourth"}
        >
          Fourth
        </DropdownItem>
      </DropdownItems>
    </Dropdown>
  )
}
```

#### Dropdown

It's just the global wrapper owning the reachui / Menu element

#### DropdownLabel

It's the information that is displayed on the screen when a value is selected (or not). It owns a `placeholder` value which will be displayed in case we don't have any children (children is an optional string here) => this means that we could use optional chaining without taking care of the fallback value in the children props

#### DropdownItems

It's a mandatory component and it's basically the one that will own the different `DropdownItem`.

#### DropdownItem

It's a simple item in the `DropdownItems` list and owns two props: `onSelect` and `selected`.

_Why not to manage the state of the selected value in the parent?_

We don't know the shape of a selected value: it can be an int, a string or even an object. I think it's up to the end user to manage their own data types on their side.

#### Screenshots

dropdown


mfrachet:
I agree with you concerning the combobox. Now, I'm just wondering about the need of adding a `selected` state to the different items, since it's not a combo nor a select, it's a simple dropdown that triggers an action a specific click.

I'm wondering about removing the `selected` props of the item here 🤔 . What do you think of this?

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+    story => (
+      <StoryUtils.Container>
+        <StoryUtils.Stack>
+          <div
+            css={(theme: Theme) => ({
+              width: "300px",
+              [theme.mediaQueries.phablet]: {
+                width: "700px",
+              },
+            })}
```
mfrachet, line 22: ℹ️ This allows to test responsiveness on the dropdown which can be a problem

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <Dropdown>
+      <DropdownLabel placeholder="Placeholder for the dropdown">
+        {selected}
+      </DropdownLabel>
+      <DropdownItems>
+        {items.map(item => (
+          <DropdownItem
+            key={item}
+            onSelect={() => undefined}
+            selected={selected === item}
```
mfrachet, line 95: ❓ Is this really necessary? A dropdown is supposed to be a list of action right? I'm wnodering about the need of having a selected state here 🤔

mfrachet:
+ I'm having some weird trouble during the checks with the navigation components and some other, 🤔 I don't really know what's going on there

src/components/Dropdown/styles.tsx
```diff
@@ -0,0 +1,92 @@
...
+to {
+  transform: translate(0, 0) perspective(1000px) rotateX(0);
+}
+`
+
+export const dropdownCss = (theme: Theme) => ({
+  background: theme.colors.primaryBackground,
+  border: `1px solid ${theme.colors.grey[10]}`,
+  borderRadius: theme.radii[3],
+  boxShadow: `0 5px 10px 3px rgba(0, 0, 0, 0.1)`,
```
Elanhant, line 14: Hm, do you mind trying out our theme shadows?

src/components/Dropdown/styles.tsx
```diff
@@ -0,0 +1,92 @@
...
+  width: "80vw",
+  transform: `translate(0, 0) perspective(1000px) rotateX(-35deg)`,
+  animation: `${enter} 0.5s ease forwards`,
+  transformOrigin: "top center",
+  overflow: "auto",
+  maxHeight: "20rem",
+
+  "[data-reach-menu-item]": {
```
Elanhant, line 22: Is it possible to pass `css` prop to `MenuItem` instead?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+} from "@reach/menu-button"
+import { MdCheck, MdKeyboardArrowDown } from "react-icons/md"
+import { dropdownCss, dropdownLabelCss } from "./styles"
+
+export const Dropdown: React.FC = ({ children }) => <Menu>{children}</Menu>
+
+// Dropdown Label
+export interface DropdownLabelProps extends Omit<MenuButtonProps, "children"> {
+  children?: string
+  placeholder: string
```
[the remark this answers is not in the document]
mfrachet: That was for the search stuff, you're right, I can switch it back 😊

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+  placeholder,
+  ...props
+}) => (
+  <MenuButton {...props} css={dropdownLabelCss}>
+    <span>{children ? children : placeholder}</span> <MdKeyboardArrowDown />
+  </MenuButton>
+)
+
+// DropdownItems
+export const DropdownItems: React.FC<MenuListProps> = props => {
```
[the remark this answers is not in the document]
mfrachet: I'm not sure to get your comment :/. I'm defining `DropdownItems` as a functional component of `MenuListItemProps` here

src/components/Dropdown/styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+  border: `1px solid ${theme.colors.grey[30]}`,
+  borderRadius: theme.radii[2],
+  color: theme.colors.grey[90],
+  fontSize: theme.fontSizes[2],
+  padding: `${theme.space[2]} ${theme.space[4]}`,
+  width: "100%",
+
+  "&:focus": {
+    borderColor: theme.colors.purple[40],
+    boxShadow: `0 0 0 3px ${theme.colors.purple[20]}`,
```
mfrachet, line 54: ℹ️ I didn't find an appropriate shadow in theme :/ Should be add one for the input outlines?

src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
Elanhant: I did feel that my comment might be a bit confusing 😓 Let me try to rephrase that.

So imagine we ship this as it is, and you want to use this `DropdownItems`component in a TypeScript project; but you also want to customize its behaviour via some prop e.g. `variant`. 

So you start writing your wrapper component:
```typescript jsx
import { DropdownItems } from "gatsby-interface"

function CustomDropdownItems({ variant, ...props }) {
  // do something with variant
  // ...

  return <DropdownItems {...props} />
}
```

But now you need to define a type for `CustomDropdownItems` props. Intuitively (from my point of view, at least), you try to do this:
```typescript
import { DropdownItems, DropdownItemsProps } from "gatsby-interface"

type CustomDropdownItemsProps = DropdownItemsProps & {
  variant: "FOO" | "BAR"
}
```

But `DropdownItemsProps` is not available, so you'll have to go view `index.d.ts` file in gatsby-interface to check what are the actual props for `DropdownItems`; and then import `MenuListProps` from `@reach/menu` and use it instead. 

And this leads to at least two problems:
1. Less clarity — if other people were to work on the same component, they would see this:
    ``` typescript jsx
    import { DropdownItems } from "gatsby-interface"
    import { MenuListProps } from "@reach/menu"

    type CustomDropdownItemsProps = MenuListProps & {
      variant: "FOO" | "BAR"
    }
  
    function CustomDropdownItems({ variant, ...props }: CustomDropdownItemsProps) {
      // do something with variant
      // ...
    
      return <DropdownItems {...props} />
    }
    ```
    This might look confusing — why do we render `DropdownItems`, but use props from `@reach/menu`?
1. Exposing implementation details — people would have to know that they need to rely on `@reach/menu` for some of the types. But what if we change `DropdownItems` to accept some new prop (or, less probably, rewrite it completely to not use ReachUI?).

Because of this, I encourage **always** declaring and exporting a props type for each component exposed by a library:
```typescript jsx
import { MenuList, MenuListProps } from "@reach/menu"

export type DropdownItemsProps = MenuListProps

export const DropdownItems: React.FC<DropdownItemsProps> = props => {
  return <MenuList {...props} css={dropdownCss} />
}
```

I hope this explains my position more clearly :)

src/components/Dropdown/styles.tsx
Elanhant: Not sure to be honest, let's ask @fk

src/components/Dropdown/index.ts
```diff
@@ -0,0 +1,6 @@
+export {
```
Elanhant, line 1: Please use wildcard export so that types are also available for library consumers:
```typescript
export * from "./Dropdown" 
```

src/components/Dropdown/styles.tsx
```diff
@@ -0,0 +1,73 @@
+import { Theme } from "../../theme"
```
Elanhant, line 1: Could you please rename this file to `Dropdown.styles.tsx` to be consistent with other components?

Elanhant:
> * I'm having some weird trouble during the checks with the navigation components and some other, 🤔 I don't really know what's going on there

This was due to `styles.tsx` importing `Theme` type from `../..` instead of `../../theme`; `../..` references `src/index.ts` which imports three remaining JS files. This file is explicitly excluded when building type definitions for the bundle, but once you import it in any other file, it gets included again.

I have fixed that in [this commit](https://github.com/gatsby-inc/gatsby-interface/pull/275/commits/c66fa5d66375981bb33cd10befdaf193adb8be34).

src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
mfrachet: Super clear! Thank you for this 😊

mfrachet:
Thanks @Elanhant ! Just made the modifications 😊

src/components/Dropdown/styles.tsx
fk: Phew, good question. There _should_ be one ;)

PR #276, opened 2020-04-06
Styled Modals (Part 2)

Elanhant:
Several things going on here:

1. Added `StyledModalActions` component to be used as part of `StyledModal` organism
1. Added `StickyObserver` component that adds a "lip" shadow to sticky elements when they get "stuck"
1. Adds `StyledPanel` organism; currently we only use panels for raw logs view in Cloud, but they can be a powerful tool for other cases as well. Because we don't have other examples of a panel usage, I've relied on my previous experience when prototyping styled panels. 

Main differences of `StyledPanel` compared to `StyledModal`:
* Header and actions are sticky — panels generally take up all of the available vertical screen space, which means that they tend to have scrollable content. This naturally favours sticky headers and footers, as they contain important context/actions which should be visible at all times.
* "Body section" component instead of "body" — panels allow for more diverse content in them, such as lists, forms or even full-fledged widgets (e.g. messaging). If we were to have a "body" component like `StyledModalBody`, with predefined padding and other styles, we'd have to constantly reset those styles if we wanted to render something that does not need them. Our current usage (raw logs) is a great example — there's no need for padding there, as the logs view has its own spacing and background. Screen Shot 2020-04-06 at 2 44 04 PM So instead there is a `StyledPanelBodySection` component, which would apply default styles; also, `StyledPanelActions`, unlike its modal counterpart, must be rendered outside of "body section", as it takes care of spacing on its own. 

[Demo of StyledPanel is available in Chromatic](https://5d4c6885a869c600201d3c8b-qdcyjzpmfz.chromatic.com/iframe.html?id=modal-styledpanel--basic). Note that this demo is displaying a panel outside of "modal dialog" flow, it is just the panel component itself, no overlay, no width restriction.


src/components/Modal/StyledModal.stories.tsx
```diff
@@ -52,7 +54,13 @@ const LONG_TEXT = `Lorem ipsum dolor sit amet consectetur adipisicing elit. Quo
 export const Basic = () => (
   <StyledModal>
     <StyledModalHeader>Header</StyledModalHeader>
-    <StyledModalBody>{LONG_TEXT}</StyledModalBody>
+    <StyledModalBody>
+      {LONG_TEXT}
+      <StyledModalActions>
```
mfrachet, line 59: ❓ I'm wondering about the semantics of these: shouldn't `StyledModalActions` be sibling to `StyledModalHeader` and `StyledModalBody ` instead of a child of `StyledModalBody` ?

_EDIT: I've seen that in the Panel, they are well siblings but not there_

src/components/Modal/StyledPanel.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+          styles={(theme: Theme) => [
+            {
+              body: { background: theme.colors.grey[20] },
+            },
+            isChromatic() && {
+              // Make animations instant so that Chromatic can take proper snapshots
+              "*, :before, :after": {
+                animationDuration: `0s !important`,
+                animationDelay: `0s !important`,
+              },
```
mfrachet, line 39: awesome trick 👍 and thanks for the heads up, I've forgot about that 😅

src/components/Modal/StyledModal.stories.tsx
Elanhant: You know, I don't think I even considered making `StyledModalActions` a sibling of body and header! Probably because its Cloud counterpart is always being used inside a body. 

I don't see any reason not to render it as a sibling, will try it out.

Elanhant: Actually, I do see a reason against "sibling" position: if a modal/panel contains a form and the "actions" container has a submit button. But this can be fixed with something like this:
```jsx
<StyledModal>
  <StyledModalHeader>Form</StyledModalHeader>
  <form css={{ margin: 0, padding: 0 }}>
    <StyledModalBody>
      ...
    </StyledModalBody>
    <StyledModalActions>
      ...
    </StyledModalActions>
  </form>
</StyledModal>
```

Elanhant: Hm, it's not as trivial to change as I hoped it would be — `StyledModalBody` has border radius to get those rounded corners, and if we make `StyledModalActions` a sibling we'll need to take care of this somehow :( 

If it's ok with you, I think we should ship this and think about possible solutions later.

PR #279, opened 2020-04-08
chore: add an util to list prop variations in a sandbox

Elanhant:
Adds a new Storybook util, `sandboxWithPropVariations`, that allows to display a component with different values for a selected prop via Knobs add-on.

To use this util function, you'll have to update your "Sandbox" story. For example, if you had something like this:
```jsx
const VARIANTS: MyComponentVariant[] = ['foo', 'bar']

export const Sandbox = () => (
  <MyComponent
    children={text('children', 'Lorem ipsum')}
    variant={radios('variant', radioKnobOptions(VARIANTS), 'foo')}
  />
)
```
then you'll need to change it to this:
```jsx
const VARIANTS: MyComponentVariant[] = ['foo', 'bar']

export const Sandbox = () => sandboxWithPropVariations(propVariations => (
  <MyComponent
    children={text('children', 'Lorem ipsum')}
    variant={radios('variant', radioKnobOptions(VARIANTS), 'foo')}
    {...propVariations}
  />
), {
  variant: VARIANTS,
})
```

You can specify possible variations for multiple props and then select which of them do you want to see in Storybook Knobs add-on panel:
Screen Shot 2020-04-08 at 10 19 55 AM
Screen Shot 2020-04-08 at 10 22 15 AM

### DIFF
--- a/src/components/Notification/Notification.stories.tsx
+++ b/src/components/Notification/Notification.stories.tsx
@@ -10,7 +10,7 @@ import {
   NotificationVariant,
   NotificationTone,
 } from "."
-import { StoryUtils } from "../../utils/storybook"
+import { StoryUtils, sandboxWithPropVariations } from "../../utils/storybook"
 import { Button } from "../Button"
 import { radioKnobOptions } from "../../utils/storybook/knobs"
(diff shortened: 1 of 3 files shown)


fk:
YAAAAS! 🙏

PR #281, opened 2020-04-08
chore: group some of the stories

Elanhant:
Gathers form, icon and modal stories under top level groups:
Screen Shot 2020-04-08 at 3 49 52 PM


Elanhant:
> That is absolutely AWESOME! love it! And it makes me think of this article which I think, is pretty good in what we may want to achieve!
> 
> https://www.designsystems.com/stack-mirroring-designing-for-code-and-coding-for-design/

Thanks for the link, I'll take a look 👍

fk:
<3 <3 <3

PR #284, opened 2020-04-13
Form refactoring

Elanhant:
A lot of stuff going on here, here's the breakdown:
1. Introduce `layout` prop to all "block" fields (such as `InputFieldBlock` or `InputConnectedField`), with the exception of `CheckboxField` (not to be confused with `CheckboxGroupField`). 
This prop allows to display field labels next to the corresponding controls, not above them ([see example here](https://5d4c6885a869c600201d3c8b-neplatredm.chromatic.com/?path=/story/form-styled-blocks-inputfieldblock--sandbox). This warranted some of the other changes listed here.
1. Field primitives (such as `InputField`, `InputFieldControl`, `InputFieldLabel`) no longer take care of their margins. Instead, there is now a `FormFieldContainer` component (and its hook counterpart, `useFormFieldContainerProps`), which is basically a CSS grid to hold 4 essential elements of a field: label, control, error and hint. `FormFieldContainer` allows to easily switch between vertical and horizontal layout with a bit of CSS without any changes in the DOM tree. 
1. Because of relying on CSS Grid, we had to abandon using `<fieldset>` and `<legend>` in radio button and checkbox group fields — Chromium browsers have a long-standing issue with applying `display: grid` or `display: flex` to `<fieldset>` elements (and `<legend>` can only ever be a direct child of a fieldset, so we wouldn't be able to work around this with some wrapper `<div>` ).
  Instead, we are going to use `role="group"` together with `aria-labelledby` [as recommended by W3C](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria).
1. `CheckboxGroupField` and `RadioButtonField` used to support their own `layout` prop, which was responsible for **options** layout; I have renamed it to `optionsDirection` and its supported values to `row` and `column`, similar to `flex-direction`.
1. `CheckboxGroupField` has got the same changes I did for `RadioButtonField` in #283 
1. `FormGroupFieldProvider ` no longer accepts a `variant` prop — this prop was specific to `RadioButtonField`, and therefore was moved there (and removed from context).

There are also a lot of changes in stories for form components, but I suggest to not pay too much attention to those, since I'm probably going to rewrite them completely.


Elanhant:
@mfrachet I think @fk confirmed that we need to support horizontal layout when we were discussing Site Details form. 

I am not sure if it is feasible to rely on styled primitives for this on per case basis, instead of baking it into props, but I can look into it.

fk:
> @mfrachet I think @fk confirmed that we need to support horizontal layout when we were discussing Site Details form.

Top-aligned labels are/vertical layout is the preferred layout, but 💯yes, we want to support both (and not only because of our "edit-in-place" scenario which sparked this conversation).

I'm running behind, so for now let me c&p some other smart people:

- (from https://uxdesign.cc/design-better-forms-96fadca0f49c (emphasis mine)):  
    
  > Users complete top aligned labeled forms at a much higher rate than left aligned labels. Top aligned labels also translate well on mobile. **However, consider using left aligned labels for large data-set entry with variable optionality because they are easier to scan together, they reduce height, and prompt more consideration than top aligned labels.**. 
    
  image
- from the 💜 classic https://www.lukew.com/ff/entry.asp?504 (emphasis mine):  

  > Left-aligned forms are the slowest of the three to complete because of the number of eye fixations they require to parse. **However, for forms with lots of optional fields or unfamiliar data (like preferences dialogs or advanced settings), they allow users to effectively scan labels. In fact, if you want users to slow down and consider each input in a form more carefully, left-aligned labels are a good way to go.**

PR #286, opened 2020-04-15
BaseNavigation update

greglobinski:
## Purpose

Resolves the issue we encountered working on willit.build 

- story: https://app.clubhouse.io/gatsbyjs/story/4165/mobile-navigation-overlay-does-not-disappear-after-switching-to-a-new-page 
- pr: https://github.com/gatsby-inc/willit.build/pull/108 (with screencasts)

## Changes

to fix the issue we:
- extend `BaseNavigation` by adding `linkClosesMobileNav` prop
- if the prop is `true` we trigger `setIsMobileNavOpen` to set  `isMobileNavOpen` value to `false`  every time any of nav links is clicked.


Elanhant:
> This looks great to me, though I wonder if we need `linkClosesMobileNav`—my intuition is that clicking a link should always close the nav, I'm not sure why you would ever want this to be `false`.
> 
> Would also be good to hear from @Elanhant on this one :)

Good point, I haven't thought of that! Yeah I think it is reasonable to always close the nav on link click.

PR #288, opened 2020-04-15
feat: Navigation rewritten to TypeScript

Elanhant:
As the title says, this is a rewrite of `Navigation` component to TypeScript. There shouldn't be any changes in how it works though.


fk:
✨

PR #289, opened 2020-04-16
feat: Spacer component

Elanhant:
This is a prototype for a `Spacer` component that allows to display an empty space (implemented as a `<div />`) with a size supported by `theme.space`.

* supports horizontal and vertical (default) direction
* supports specifying size for a theme breakpoint via `responsiveSize` 
* supports specifying direction for a theme breakpoint via `responsiveDirection` 
* I left `className` prop as an escape hatch in case some adjustment is needed, though hopefully there won't be many cases for it 

[Demo available in Chromatic](https://5d4c6885a869c600201d3c8b-ogsyoqrsen.chromatic.com/?path=/story/spacer--basic).


src/components/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const defaultSizeProperty = direction === `horizontal` ? `width` : `height`
+
+  return (
+    <div
+      css={[
+        {
+          minHeight: `1px`,
+          minWidth: `1px`,
+        },
```
mfrachet, line 72: ❓ why are these necessary?

Elanhant: Hm, spacer didn't work properly for me without those but now it seems ok, at least in Storybook 🤔 Will remove them.

PR #290, opened 2020-04-16
Adding ProgressBar

mfrachet:
Adding a progress indicator to the library. The background color changes depending on the progression

## API

```jsx
<ProgressBar value={3} max={14} id="unique-id-for-describedby">
  <strong>3</strong> days remaining
</ProgressBar>
```

## Screenshots

progress bar

## Notes

I was thinking of isolating the progress bar (the colored indicator) in a specific component but I think that it would lose some contextual information not to own its label. I _feel_ that it's more semantic to wrap the visual representation and the data we want to share at the same level 🤔


src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  if (progression < 66) {
+    return theme.colors.orange[50]
+  }
+
+  return theme.colors.green[50]
+}
+
+const progressBarCss = (_theme: Theme): CSSObject => ({
```
Elanhant, line 20: You can use our local `ThemeCss` type instead:
```typescript
const progressBarCss: ThemeCss = _theme => ({
  // ....
})
```

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export const ProgressBar: React.FC<ProgressBarProps> = ({
+  max,
+  min = 0,
+  value,
+  children,
+  id,
+}) => {
+  const progression = (value / max) * 100
```
Elanhant, line 70: We need to normalise progress value to keep it between `0` and `100`. Otherwise the bar will overflow if we receive some overaged value :)

src/index.ts
```diff
@@ -72,3 +72,4 @@ export * from "./components/StepIndicator"
 export * from "./components/DensityProvider"
 export * from "./components/StickyObserver"
 export * from "./components/Dropdown"
+export * from "./components/ProgressBar"
```
Elanhant, line 75: We should make the same change in `index-ts-only.ts` in the root dir.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const progressBarCss = (_theme: Theme): CSSObject => ({
+  display: "flex",
+  flexDirection: "row-reverse",
+  alignItems: "center",
+})
+
+const progressCss = (theme: Theme, progression: number): CSSObject => ({
+  width: "50px",
```
Elanhant, line 27: Let's not restrict the width of a progress bar, this should be the responsibility of whatever component is going to use it.

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  min = 0,
+  value,
+  children,
+  id,
+}) => {
+  const progression = (value / max) * 100
+
+  return (
+    <div css={progressBarCss}>
+      <p css={progressBarLabelCss} id={id}>
```
Elanhant, line 74: Hm, I don't think we should include associated label in the `ProgressBar` component. Quite often these progress bars are used in places where we cannot display the label in the same way (or at all).

Better to keep this `ProgressBar` atomic and allow passing `aria-describedby` prop to link it to some label:
```typescript
export type ProgressBarProps {
  max: number
  min?: number
  value: number
  'aria-describedby': string
}

// also don't forget to remove React.FC in this case to indicate that children are not allowed 
```

We can always have a `LabelledProgressBar` component sitting next to this one if we want 🙂

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  strong: {
+    color: theme.colors.black,
+  },
+})
+
+export interface ProgressBarProps {
+  max: number
+  min?: number
+  value: number
+  id: string
```
Elanhant, line 60: We should also allow passing custom color for the bar.

nachiketkumar:
LGTM in storybook :)

src/components/ProgressBar/ProgressBar.tsx
mfrachet: Let's leave the label display responsibility to the host app :)

src/components/ProgressBar/ProgressBar.tsx
mfrachet: I've passed a custom (optional) function receiving the theme and the current progression (in %) so that the end user can tweak it the way they want

fk:
<3

Issue #47, opened 2019-09-10
Get rid of intermediate step to generate declarations once all source code is rewritten in TypeScript


Elanhant:
- [x] Remove `./index-ts-only.ts` as at this point `./src/index.ts` should contain only TypeScript imports.
- [x] Either remove or update `tsconfig.build-types.json` to include `./src/index.ts`
- [x] Remove `copy` and `del` steps from `rollup.config.json` that are used to rename `index-ts-only.d.ts` to `index.d.ts`.

PR #295, opened 2020-04-20
Only allow TypeScript in the library code

Elanhant:
~**DO NOT MERGE:** I need to ensure that .com will not fail after these changes. I will merge this PR myself when I am confident that it won't break anything.~

This PR switches `allowJs` option in TypeScript config to `false` and contains several changes necessary for that to work
* `FileUpload` component has been removed — confirmed with @gillkyle that it was supposed to be used for resume upload, but we ended up using plain `type="file"` input. This also means that `gatsby-interface` will not depend on `filestack-react` anymore (which should also solve a common problem with `filestack-js` mentioned in the [Usage doc](https://gatsby-interface.netlify.app/?path=/docs/about-gatsby-interface-usage--page))
* New stories have been added for `Link` component instead of the outdated JS ones.
*  `index-ts-only.ts` is no more, `src/index.ts` now is the entry point both for the library code and for the types (solves #47).

I have also removed legacy button components (`PrimaryButton` etc) which are not used by .com anyway.


fk:
> I have also removed legacy button components (PrimaryButton etc) which are not used by .com anyway.

💜 Was wondering about these! 🙏

mfrachet:
That's amazing! -950 lines is awesome! Thanks 🎉

PR #296, opened 2020-04-20
feat: make `<Button>` use `fonts.sans` by default, add prop `textVariant` to use them w/ "Futura PT"

fk:
image

- Use `maxHeight` to define button heights (as opposed to fiddling around with `font-size`, `line-height`, and vertical `padding`). This _doesn't_ significantly change the height of the buttons, but only evens them out — with a slight adjustment to the `XL` size being the exception:
  - `S`: `~28px` -> `28px`
  - `M`: `~36px` -> `36px`
  - `L`: `~41px` -> `40px`
  - `XL`: `~52px` -> `48px`
- Analog to what we are doing for `<Badge>`, add a prop `textVariant`:
  - Defaults to `DEFAULT` 😅, which renders buttons using `theme.fonts.body` (/`.sans`).
  - For those cases where our buttons should reflect the Gatsby brand, `BRAND` brings back "Futura PT" at the previously defined, slightly bigger font-sizes (to visually match the size of `theme.fonts.sans`). This hopefully makes transitioning to the new `<Button>` defaults as easy as possbile. 🤞
  - Add knobs for `textVariant`.


fk:
> is there any specific reason to use `maxHeight`

Ha you scared me for a bit there — "Why `maxHeight`, did I use `max`?" … :)
(it's `minHeight` :))

Ideally something like this will never happen, but with `height` …

Bildschirmfoto 2020-04-24 um 01 12 57

vs `min-height`:

Bildschirmfoto 2020-04-24 um 01 13 17

`min-height` is a easy way out of fiddling around with vertical padding — it only works along with `inline-flex` and `align-items: center`, and particulary … OK ;-) because our buttons aren't all caps. The approach (as opposed to the "fiddling around with `font-size`, `line-height`, and vertical `padding`") def. also has problems — one is that the current vertical padding (which really is just there as a buffer for exactly this (edge?) case) doesn't really hold up when that `min-height` is exceeded, which shows in the `min-height` screenshot above.

fk:
Writing the above just made me wonder how others do it (hey, I'm old! ;)), and it's a mix — [Material](https://material.io/components/buttons) also seems to use `inline-flex`, but along `height`, others do padding.

Re:`flexbox` for `<button>`, I also stumbled upon https://stackoverflow.com/a/35466231. Looking at the open issues listed in https://github.com/philipwalton/flexbugs#flexbug-9, the most crucial was https://bugs.chromium.org/p/chromium/issues/detail?id=700029, which has been fixed, and as far as I understand in v83. However, the codepen linked in https://chromestatus.com/feature/5213452823953408 is working in my Desktop Chrome v80.something, but it's late — I might be missing something.

fk:
I hope the above made sense — merging because `display: inline-flex` was in use before, and IMO the other changes are worth making right now.

PR #297, opened 2020-04-23
Updates to ReachUI-based components

Elanhant:
* Update all `@reach/*` dependencies to their latest versions
* Rename `Dropdown` to `DropdownMenu` to better convey its purpose (so that we don't use it as a form input 😉) and expose more subcomponents based on `@reach/menu-button` exports.
* Fix `Tooltip`'s pointer rendering which breaks with the newer versions of `@reach/tooltip`
* Expose `useComboboxContext` (provided by `@reach/combobox`) hook from `Combobox` (and rename our own hook of the same name to `useComboboxCustomContext`)


src/components/Combobox/Combobox.tsx
```diff
@@ -210,3 +211,7 @@ export type ComboboxOptionTextProps = {}
 export function ComboboxOptionText(props: ComboboxOptionTextProps) {
   return <ReachComboboxOptionText {...props} />
 }
+
+export function useComboboxContext() {
+  return useReachComboboxContext()
+}
```
mfrachet, line 217: I'm not sure to get why we would need it to be public?

Elanhant: Because `@reach/combobox` made it public 🙂 
Until we have a clear understanding on how we are going to use `Combobox`, it is better to mirror ReachUI's API as much as possible. Then, once we start noticing some patterns in our usages of this component, we can start being more restrictive.

PR #298, opened 2020-04-24
Navigation dropdowns  with dynamic width

greglobinski:
## Purpose

- The PR removes fixed `width` `450px` from `Navigation > Dropdown`. 
- Every dropdown item gets its `width` based on its content. 
- There is a supplemental PR with viewport collision detection > https://github.com/gatsby-inc/gatsby-interface/pull/299. 
- **The PR can't be merged before the supplemental PR mentioned above is merged into it.**  

Navigation - With Wide Dropdowns ⋅ Storybook ⋅ Storybook (1)

Note: this PR is a new version  (in TypeScript) of outdated PR https://github.com/gatsby-inc/gatsby-interface/pull/189


src/components/BaseNavigation/BaseNavigation.styles.ts
```diff
@@ -74,19 +74,21 @@ const baseNavigationItemInvertedCss: ThemeCss = theme => ({
...
   position: `absolute`,
-  top: `95%`,
-  left: 0,
-  margin: 0,
-  padding: `${theme.space[4]} 0`,
+  top: `100%`,
+  left: `50%`,
+  transform: `translateX(-50%)`,
+
+  ul: {
```
Elanhant, line 83: Let's extract these `ul` styles into a separate const and use it directly as a `css` prop.

greglobinski: Does it mean we should expose `ulCss` prop  (or something like this) on `BaseNavigationDropdown` to let overwrite the default styles of it when the `BaseNavigationDropdown` is restyled in `NavigationDropdown`?

Elanhant: Sorry for taking so long to respond. 

Ideally, we should split up `BaseNavigationDropdown` into something like `BaseNavigationDropdownPopover` (the outer `<div>`) and `BaseNavigationDropdownItems` (the `<ul>` element) and add their support to the `BaseNavigationContext`. In my opinion, that would be the preferred solution here. Alternatively, we could indeed introduce some new prop like `dropdownItemsClassName` and use [`<ClassNames>`](https://emotion.sh/docs/class-names) component from Emotion to pass it.

I also think that we might have overcomplicated both Navigation/BaseNavigation components, and we should probably revise them at some point. I am already struggling with reading through the code 😅 But this is by no means in the scope of this PR.

greglobinski: > I also think that we might have overcomplicated both Navigation/BaseNavigation components, and we should probably revise them at some point. 

:100:  Especially that there is still a room to improve the component.

For example the dropdowns' behaviours when user navigate using `tab` key. This was a part of the original PR https://github.com/gatsby-inc/gatsby-interface/pull/189, but we agreed then than that the implementation was not the way we should do it so I left it unchanged for now. 

And because of that I went the second option - added `dropdownListClassName` prop to `BaseNavigationDropdown` and use `ClassNames` in `NavigationDropdown`

PR #308, opened 2020-04-26
fix(Notification): `Link` margin-left

### LINKED ISSUE #291: fix: `Notification` `Link` needs margin
image

fk:
Fixes #291:

- add `margin-right` to `<NotificationContent>` _only if there's a `<Link>`_
- the latter is needed because we need `margin-left: auto` to right-align `<Link>` in case of the `<NotificationDismissButton>` also being present, to prevent this: 
image

---

Before/after:

image

image

https://deploy-preview-308--gatsby-interface.netlify.app/?path=/story/notification--sandbox

---

It's not all good though:

- `<Link>` needs some love for small screens — probably displaying it in a dedicated row below the content is the way to go. Currently we display two (content, link), potentially three (dismiss button) columns, which isn't ideal at all.
- Wondering if we should deprecate `<Link>`, or at least think about
  - whether `<Link>` and `<NotificationDismissButton>` should be mutually exclusive
  - two content areas, one mandatory, one optional


Elanhant:
> <Link> needs some love for small screens — probably displaying it in a dedicated row below the content is the way to go. 

Yeah, sounds good to me!

PR #321, opened 2020-04-30
Adding Trialing (professional trial) as a plan

mfrachet:
We need to add the trialing / professional plan as a badge 

This badge doesn't have an icon and is in green

Screenshot 2020-04-30 at 18 26 29

### DIFF
--- a/src/components/PlanIndicator/PlanIndicator.tsx
+++ b/src/components/PlanIndicator/PlanIndicator.tsx
@@ -37,11 +37,18 @@ const planTypeEnterpriseCss: ThemeCss = theme => ({
   color: theme.colors.white,
 })
 
+const planTypeTrialingCss: ThemeCss = theme => ({
+  borderColor: theme.colors.green[10],
+  backgroundColor: theme.colors.green[10],
+  color: theme.colors.green[70],
+})
+
 const planTypeCss: Record<PlanIndicatorPlanType, ThemeCss> = {
   FREE: planTypeFreeCss,
   PROFESSIONAL: planTypeProfessionalCss,
   BUSINESS: planTypeBusinessCss,
   ENTERPRISE: planTypeEnterpriseCss,
+  TRIALING: planTypeTrialingCss,
 }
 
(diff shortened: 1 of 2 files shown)


src/components/PlanIndicator/PlanIndicator.tsx
```diff
@@ -37,32 +37,45 @@ const planTypeEnterpriseCss: ThemeCss = theme => ({
...
   PROFESSIONAL: `Professional`,
   BUSINESS: `Business`,
   ENTERPRISE: `Enterprise`,
+  TRIALING: "Professional Trial",
 }
 
-const planTypeIcons: Record<PlanIndicatorPlanType, React.ComponentType> = {
+const planTypeIcons: Record<
+  PlanIndicatorPlanType,
+  React.ComponentType | undefined
```
Elanhant, line 64: I think it's better to use `null` instead of `undefined` since we do "define" a `TRIALING` icon 🙂

src/components/PlanIndicator/PlanIndicator.tsx
```diff
@@ -37,32 +37,45 @@ const planTypeEnterpriseCss: ThemeCss = theme => ({
...
   PROFESSIONAL: `Professional`,
   BUSINESS: `Business`,
   ENTERPRISE: `Enterprise`,
+  TRIALING: "Professional Trial",
 }
 
-const planTypeIcons: Record<PlanIndicatorPlanType, React.ComponentType> = {
+const planTypeIcons: Record<
+  PlanIndicatorPlanType,
+  React.ComponentType | null
```
Elanhant, line 64: Oops, my bad! `null` breaks the build script. What if you use `BadgeProps['Icon']`?

mfrachet: null is not accepted as part of the type 😅

mfrachet: However, using BadgeProps and `undefined` looks working :)

PR #327, opened 2020-05-01
chore: Realign `SidebarNav` subitems to save horizontal space

fk:
Saves a bit of horizontal space, and (subjectively ;)) feels better — before/after:

image

### DIFF
--- a/src/components/SidebarNav/SidebarNav.tsx
+++ b/src/components/SidebarNav/SidebarNav.tsx
@@ -153,8 +153,12 @@ function SidebarNavListItem({
               key={subItem.to}
               css={(theme: Theme) => [
                 {
-                  padding: `${theme.space[3]} ${theme.space[5]}`,
+                  paddingTop: theme.space[3],
+                  paddingRight: theme.space[5],
+                  paddingBottom: theme.space[3],
+                  paddingLeft: theme.space[6],
                   marginBottom: `0`,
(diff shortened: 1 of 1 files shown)


src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -153,8 +153,12 @@ function SidebarNavListItem({
               key={subItem.to}
               css={(theme: Theme) => [
                 {
-                  padding: `${theme.space[3]} ${theme.space[5]}`,
+                  paddingTop: theme.space[3],
+                  paddingRight: theme.space[5],
+                  paddingBottom: theme.space[3],
+                  paddingLeft: `calc(${theme.space[6]} + 1px)`,
```
mfrachet, line 159: Mmh... While I really appreciate the move, I don't feel super confortable about computing the token value: that can break consistency across the app because it would create a new set of space values that we don't explicitly know (or can use explicitely). 

Curious about other people thoughts on that.

fk: Tokens should contain the most common values and units we use. While we should stick to them whenever possible, it's OK to break the system once in a while—especially if it is to implement it.

In this case, the `calc` compensates for two things: Using `border-left` which consumes space—as opposed to e. g. an absolute positioned `after` with a `left: -1px`—and "fake-centering" the 1px-wide line below the 20px-width icons. Happy to refactor!

fk:
No worries Marvin!

Issue #322, opened 2020-04-30
Add „Tabs“ component


fk:

fk:
Added in #329, thank you @Elanhant! <3

PR #329, opened 2020-05-04
Tabs component

Elanhant:
Moves `Tabs` component from Cloud to gatsby-interface (#322).

We are going to expose `tabsStyles` for the time being since our Site/Org tabs are implemented not as actual tabs, but rather as a nav menu.  

Preview can be found here: https://5d4c6885a869c600201d3c8b-qtoydhgfdt.chromatic.com/?path=/story/tabs--basic


fk:
<3

PR #332, opened 2020-05-11
chore: Remove `SidebarNav variant="FULL"`

### LINKED ISSUE #326: SidebarNav: Consider removing the `FULL` variant
https://github.com/gatsby-inc/gatsby-interface/blob/8038a56acf5eda42ebe72c40c048dde080b4695b/src/components/SidebarNav/SidebarNav.tsx#L63

Should we allow setting `maxWidth` instead?
This feels very "magic number", also IMO the variant name is misleading/confusing.

fk:
…and thus all variants of the component:

- Not used in Gatsby Cloud.
- `maxWidth: 8rem` that the `FULL` variant sets on `SidebarNavList` could be set on its parent `SidebarNav` via Emotion's `css` prop or `styled` (at `10rem`, taking `SidebarNav`s `padding-left` into account).
- Remove `Variants` story, knob in `Sandbox` story.

Fixes #326

### DIFF
--- a/src/components/SidebarNav/SidebarNav.tsx
+++ b/src/components/SidebarNav/SidebarNav.tsx
@@ -42,7 +42,7 @@ export default function SidebarNav({ options, ...rest }: SidebarNavProps) {
 
 type SidebarNavListProps = JSX.IntrinsicElements["ul"]
 
-function SidebarNavList({ ...rest }: SidebarNavListProps) {
+function SidebarNavList(props: SidebarNavListProps) {
   return (
     <ul
       css={[
@@ -52,7 +52,7 @@ function SidebarNavList({ ...rest }: SidebarNavListProps) {
           padding: 0,
         },
       ]}
-      {...rest}
+      {...props}
     />
   )
 }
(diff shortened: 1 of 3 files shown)


fk:
@greglobinski Could you take a look at this? Anything I'm missing?

src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -47,11 +40,9 @@ export default function SidebarNav({
   )
 }
 
-type SidebarNavListProps = JSX.IntrinsicElements["ul"] & {
-  variant?: SidebarNavVariant
-}
+type SidebarNavListProps = JSX.IntrinsicElements["ul"]
 
-function SidebarNavList({ variant = `DEFAULT`, ...rest }: SidebarNavListProps) {
+function SidebarNavList({ ...rest }: SidebarNavListProps) {
```
greglobinski, line 45: After removing `variant` there is no `rest` and we do not need to destructure the the parameters.

fk: 👍 🙏

fk: d158baf

PR #333, opened 2020-05-20
fix: fix color contrast issues in Announcement and StepIndicator

Elanhant:
We had a couple of color contrast issues with `Announcement` and `StepIndicator` components:
* `Announcement` text color had insufficient contrast and it used a background image which prevents automated tooling from detecting the contrast issues.
* Inactive and "disabled" states of `StepIndicator` steps had insufficient contrast.

To solve these issues I did the following:
* Switched text color for `Announcement` from `teal` to `green` and removed background image in favour of rendering the SVG next to the text. This reduces the amount of space for the announcement text, but it also helps tooling like axe to properly detect color contrast and helps to avoid text overlapping with the "dots":Screen Shot 2020-05-20 at 11 46 49 AM
We can also try using [`DecorativeDots`](https://gatsby-interface.netlify.app/?path=/story/decorativedots--default) instead of this static SVG.
* `StepIndicator` now uses  `grey[60]` instead of `grey[40]` for the step label:Screen Shot 2020-05-20 at 11 50 15 AM


src/components/StepIndicator/StepIndicatorStep.styles.ts
```diff
@@ -4,9 +4,9 @@ export const baseCss: ThemeCss = theme => ({
   display: `inline-flex`,
   alignItems: `center`,
   borderRadius: `${theme.radii[2]} ${theme.radii[2]} 0 0`,
-  background: theme.colors.secondaryBackground,
+  backgroundColor: theme.colors.grey[5],
   border: `1px solid ${theme.colors.standardLine}`,
-  color: theme.colors.grey[40],
+  color: theme.colors.grey[60],
```
fk, line 9:
```suggestion
  color: theme.colors.grey[50],
```

would give us a color contrast of `4.39`, which seems fine for this text size (`18px`), complies to AA+ / passes the Storybook a11y test?

fk: Ah no, it's not — I'm blind. ;-) Mhh, let's see if we can get a this a bit more deemphasized via `blackFade` then … :P

src/components/StepIndicator/StepIndicatorStep.styles.ts
```diff
@@ -4,9 +4,9 @@ export const baseCss: ThemeCss = theme => ({
   display: `inline-flex`,
   alignItems: `center`,
   borderRadius: `${theme.radii[2]} ${theme.radii[2]} 0 0`,
-  background: theme.colors.secondaryBackground,
+  backgroundColor: theme.colors.grey[5],
```
fk, line 7: Why switch the token here? (Noticed that `activeCss` is still using `secondaryBackground`)

Elanhant: Oops, that's a leftover from my attempts at different background colors, I'll revert it.

fk: 🙌

PR #340, opened 2020-06-11
Revised forms

Elanhant:
This is a revision of our form components internals which aims to make it easier to create new form fields with less boilerplate. 

Main changes are:
1. Two new hooks, `useAriaFormField` and `useAriaFormGroupField`: they have the same goal as `FormFieldSkeleton` and `FormGroupFieldSkeleton` — to ensure that the most basic a11y practices are applied to a form field (e.g. connections between labels, controls and errors). These can be used as a basis for whenever we need a one-off field component that is not already provided by gatsby-interface.
1. A set of new styled components for form controls and elements — these are basically just wrappers for HTML elements which should be used pretty sparingly in cases where a `*FieldBlock` component is not available. These new components are:
    * `FormError` — can be used to display a styled error message
    * `FormHint` — can be used to display a styled hint text
    * `StyledLabel` — can be used to display a styled field label
    * `StyledGroupLabel` — can be used to display a styled label for group fields 
    * `StyledCheckbox` — a styled version of `<input type="checkbox" />`; unlike most other styled primitives it also has a label component, `StyledCheckboxLabel`.
    * `StyledInput` — a styled version of `<input />`
    * `StyledRadio` — a styled version of `<input type="radio" />`; unlike most other styled primitives it also has a label component, `StyledRadioLabel`.
    * `StyledSelect` — a styled version of `<select />`
    * `StyledTextArea` — a styled version of `<textarea />`
1. All `*FieldBlock` components have been rewired to use the new hooks and components.
1. Group field blocks (`CheckboxGroupFieldBlock` and `RadioButtonFieldBlock`), as well as any other potential group fields relying on `useAriaFormGroupField`) will duplicate the error and hint in the field label (albeit hiding them visually) for better screen reader experience. For more details see [this article](https://blog.tenon.io/accessible-validation-of-checkbox-and-radiobutton-groups)

### How will these changes affect us?
There shouldn't be any noticeable changes if using `*FieldBlock` or `*ConnectedField` components. The only thing to look out for is any interactive elements in the hint/error messages. Ideally, there shouldn't be any, but if you do find something please let me know.

The biggest change (which, as I hope, is for the best) is in how we would create new one-off form components. For example, to add a combobox as a form field with styled label and error we currently have to do this:
```jsx
import {
  FormFieldContainer,
  useStyledFieldLabel,
  FormFieldSkeletonLabel,
  useStyledFieldError,
  FormFieldSkeletonError,
  FormFieldSkeleton,
} from "gatsby-interface"

function ComboboxFieldLabel({ children, size, isRequired, ...props }) {
  const styledProps = useStyledFieldLabel(children, { size, isRequired })

  return <FormFieldSkeletonLabel {...props} {...styledProps} />
}

function ComboboxFieldError({ children, ...props }) {
  const styledProps = useStyledFieldError(children)

  return <FormFieldSkeletonError {...props} {...styledProps} />
}

function ComboboxField({ hasError }) {
  return (
    <FormFieldContainer>
      <FormFieldSkeleton id="repoSelection" hasError={hasError}>
        <ComboboxFieldLabel isRequired>Select an option</ComboboxFieldLabel>

        <Combobox>
          <ComboboxInput  id="repoSelection" />
          {/* ComboboxPopover and its children */}
        </Combobox>
        <ComboboxFieldError>
          {hasError && `Failed to load options`}
        </ComboboxFieldError>
      </FormFieldSkeleton>
    </FormFieldContainer>
  )
}
```   
With this PR we should be able to reduce the boilerplate a lot by just using `FormFieldBlock`:
```jsx
import { FormFieldBlock } from "gatsby-interface"

function ComboboxField({ hasError }) {
  return (
    <FormFieldBlock
      id="repoSelection"
      label="Select an option"
      error={hasError && "Failed to load options"}
      required
    >
      {controlProps => (
        <Combobox>
          <ComboboxInput {...controlProps}  />
          {/* ComboboxPopover and its children */}
        </Combobox>
      )}
    </FormFieldBlock>
  )
}
```

Or, we we **really** need a lot of control (I hope not though), we can do this:
```jsx
import {
  useAriaFormField,
  FieldLayoutContainer,
  StyledLabel,
  FormError,
} from "gatsby-interface"

function ComboboxField({ hasError }) {
  const fieldData = useAriaFormField("repoSelection", {
    required: true,
    hasError,
  })

  return (
    <FieldLayoutContainer>
      <StyledLabel
        required={fieldData.controlProps.required}
        {...fieldData.labelProps}
      >
        Select an option
      </StyledLabel>
      <div>
        <Combobox>
          <ComboboxInput {...fieldData.controlProps}  />
          {/* ComboboxPopover and its children */}
        </Combobox>
        <FormError {...fieldData.errorProps}>Failed to load options</FormError>
      </div>
    </FieldLayoutContainer>
  )
}
```
The latter option is clearly not very different from what we do now, but the former one should help us streamline adding simpler one-off fields. In addition to that, relying on the new hooks and styled primitives should help us to avoid the current form components variation hell that I got us into 😅 We won't need `*Skeleton` components anymore, and the same probably goes for `*Field` components as well.


Elanhant:
> I won't be nitpicky in the PR, it's a lot of code

Haha thanks 😄 I do expect to see those nitpicks at some point though, they might actually point at some possible issues with this new model I'm introducing 🙂

mfrachet:
After reading the PR again, I'm seeing there's a lot of effort making the `*FieldBlock` working in multiple cases. While I think it's great to see the API getting lighter, I'm feeling there's something I don't totally get 😅 . I will ask questions so that it helps me in my pain points 😊 

- the different form elements are different at the UI / accessibility level, do we need to have new abstractions around these?
- Can't one-off field be created as part of the main application and using a composable API?

I think I'm worried the form stuff is getting complex 😅

Elanhant:
> the different form elements are different at the UI / accessibility level, do we need to have new abstractions around these?

The entire idea behind this PR was to have abstractions for the most repetitive/standard elements and functionality related to forms, such as hint/error messages and how they are connected to the control. Hence the `useAriaFormField` and `useAriaFormGroupField` as the most low-level abstractions, and `FormFieldBlock` with `FormGroupFieldBlock` to take care of labelling and auxiliary text messages.

While this PR only changes existing `*FieldBlock` (and by extension `*ConnectedField`) components, it should eventually make all of the `*Skeleton` components obsolete, as those do the same things as `useAriaFormField` but in a very distributed manner.

I hope I understood this question of yours correctly 🙂

Elanhant:
> Can't one-off field be created as part of the main application and using a composable API?

It can with our current form components, but it is not really straightforward — you'd have to use `FormFieldSkeleton` and its subcomponents, and then use lots of helpers to just reimplement the styled label, error and hint messages. 

This PR abandons the idea of "skeleton" components by replacing them with `useAriaFormField`/`useAriaFormGroupField `; it also introduces styled primitives for displaying labels, errors, hints and controls that can be combined with those hooks, instead of relying on `*Field` components.

The main advantage of this approach is that the building "blocks" for our form fields are now much less coupled, thus allowing for finer tuning of one-off fields, such as dropping in custom a11y behaviour.

mfrachet:
Thanks for taking time to answer 😄 . I'll play with these and got a better idea of their strength 😊

PR #342, opened 2020-06-17
feat: EmptyState component

Elanhant:
This PR adds `EmptyState` component with two subcomponents for call to actions: `EmptyStatePrimaryAction` and `EmptyStateSecondaryAction`.

Available for poking around in [Storybook](https://deploy-preview-342--gatsby-interface.netlify.app/?path=/story/emptystate--basic).


src/components/EmptyState/EmptyState.stories.tsx
```diff
@@ -0,0 +1,534 @@
...
+  )
+  let graphic = null
+  if (graphicType === `image`) {
+    graphic = <GraphicImage />
+  } else if (graphicType === `icon`) {
+    graphic = <MdLockOpen />
+  }
+
+  return (
+    <EmptyState
```
amberleyromo, line 71: Add `headingAs` in the sandbox story?

fk:
Oh actually I think I need to request a change — the content should be `max-width: 480px`, not the container. Container should take whatever available width, and content should be horizontally centered (and limited to 480 max-width). Does that make sense?

Elanhant:
> Oh actually I think I need to request a change — the content should be max-width: 480px, not the container. Container should take whatever available width, and content should be horizontally centered (and limited to 480 max-width). Does that make sense?

It already does work like this :) 
However, our Storybook layout prevents the container from taking all available space due to centering the story on the screen. I can play around with config to have a story which demonstrates full-width container, but in the meantime I can assure you that it works 😃

fk:
> It already does work like this :)
> I can play around with config to have a story which demonstrates full-width container, but in the meantime I can assure you that it works

wvlg63kng3kr
But hoooooowww?! 😸

No worries, don't bother tweaking Storybook stuff for this.
Don't waste time on me — trust 💯!

Elanhant:
> Don't waste time on me — trust 💯!

Oh man I feel so bad now 😞 I was sure that I implemented this component so that the container takes all available width but it turned out I was wrong! Fixed it and added a story to demonstrate this behaviour.

fk:
> Oh man I feel so bad now 😞

Shoot, I'm late, but: DON'T!!! 🤗 
https://deploy-preview-342--gatsby-interface.netlify.app/?path=/story/emptystate--fit-container-width <3

PR #344, opened 2020-06-22
Color contrast tweaks for several PlanIndicator variants

amberleyromo:
The existing color contrast for the "Free" and "Professional Trial" variants of PlanIndicator don't have sufficient contrast. To remedy that, I've changed the text colors for both.

Note that the "Free" plan was only the darkest/highest contrast orange value available, which I why I switched to a red value. To me it still seems complementary. Thoughts, @fk ?

CleanShot 2020-06-22 at 15 52 55@2x


amberleyromo:
@fk 

> I swear I thought I adjusted those at some point … 🤬 :(

I thought so too! Was surprised to find this.

fk:
> Was surprised to find this.

Together we'll find all remaining color contrast issues! 😉 

Screen_Shot_2019-02-26_at_3 38 50_PM

PR #351, opened 2020-06-30
chore(Toast): code cleanup and new stories for Toast component

Elanhant:
The main goal of this PR was to bring stories for `Toast` up to date with more recent stories, but it also updates some of the code as well by switching to object notation for styles.

However, I did not address any of the a11y concerns I have with toasts (mainly those that have interactive elements). It would warrant a separate body work 🙂


fk:
OT question: In Storybook, the `ToastProvider` container covers the UI below.
Will this be the same when the component appears in the Cloud Dashboard?

image

Elanhant:
> OT question: In Storybook, the ToastProvider container covers (as in "blocks") the UI below.
Will this be the same when the component appears in the Cloud Dashboard?

Huh this is a fun bug, I've never thought about this. It definitely can happen in Cloud Dashboard. 
On the other hand, this would force the user to actually pay some attention to all these toasts and close them 😄

Elanhant:
This PR has a ton a conflicts so I will probably close it and start anew. But before I do that, @fk — what do you think about getting rid of "error" toasts? 

From my understanding, the toast pattern should be used for messages that don't necessarily require user immediate attention (e.g. informing about some action being completed, which corresponds to our "success" toasts); errors might be better off in either a `Notification` (or similar) component or in a modal for the most serious, app-breaking errors. 

Also, I have checked Cloud code and it seems like we only use "error" toasts in 2 places, one of which is experimental. Both can (and should) be easily replaced with a modal for now.

Elanhant:
Closing in favour of #415

PR #356, opened 2020-07-16
Placeholder

mfrachet:
A proposition to add a Placeholder component

API:

```jsx
<Placeholder
        animation={fadeAnimationCss}
        css={{
          gridTemplateColumns: "auto 1fr auto",
          gridTemplateRows: "1fr 1fr 1fr",
          gridTemplateAreas: `
          "left top right"
          "left middle right"
          "left bottom right"`,
        }}
      >
        <PlaceholderBox width="40px" height="40px" css={{ gridArea: "left" }} />

        <PlaceholderBox width="70%" css={{ gridArea: "top" }} />
        <PlaceholderBox css={{ gridArea: "middle" }} />
        <PlaceholderBox width="30%" css={{ gridArea: "bottom" }} />

        <PlaceholderBox
          width="40px"
          height="40px"
          css={{ gridArea: "right" }}
        />
      </Placeholder>
```

Screenshot 2020-10-05 at 14 15 17


fk:
Would wanna look into animation possibilities a bit more, but <3 <3 <3

src/components/Placeholder/README.md
```diff
@@ -0,0 +1,9 @@
+# BaseAnchor
```
Elanhant, line 1: Readme is just a copy of readme for `BaseAnchor` :)

src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,94 @@
...
+}>({
+  space: "",
+  animation: "",
+})
+
+export interface PlaceholderProps
+  extends React.ComponentPropsWithoutRef<"div"> {
+  Left?: React.ComponentType
+  Right?: React.ComponentType
+  space?: string
```
Elanhant, line 18: Why not restrict this to theme size, similar to `<Spacer>` component?

src/components/Placeholder/Placeholder.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { fadeAnimationCss } from "./Fade"
+import { Spacer } from "../Spacer"
+
+export default {
+  title: `Placeholder`,
+  component: Placeholder,
+  subcomponents: {
+    Placeholder,
+    PlaceholderBox,
+  },
```
Elanhant, line 13: We need to apply `disableAnimationsDecorator` here, otherwise it is possible for every commit to result in a visual regression in Chromatic due to animations. See `DropdownMenu.stories.tsx` for example.

src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,94 @@
...
+  background: theme.colors.grey[30],
+  height: theme.fontSizes[3],
+})
+
+export interface PlaceholderBoxProps
+  extends React.ComponentPropsWithoutRef<"div"> {
+  width?: string
+  height?: string
+  noSpace?: boolean
+  radii?: string
```
Elanhant, line 65: Same question as for `space` above — could we restrict these to theme radii?

src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,94 @@
...
+  Right,
+  space = "0.5rem",
+  animation = "",
+  ...props
+}) => {
+  return (
+    <PlaceholderContext.Provider value={{ space, animation }}>
+      <div
+        css={{
+          display: "flex",
```
Elanhant, line 34: What if we leverage grid layout instead? This would allow us to get rid of `Left` and `Right` props and make this component incredibly flexible while still preserving spacing thanks to grid gaps?

mfrachet: I've tried to use grid in the new commits pushed :). Also, note that I tried to merge some `css` props but not sure how to handle the types here :(

Elanhant: From what I can see, you don't really need to "merge" `css` in any of the components, Emotion takes care of that as long as you allow passing `className` to the component (which you do).

mfrachet: OOh wooow! TIL about that! That's an amazing feature!

src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,75 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import React, { useContext } from "react"
```
Elanhant, line 3: To be consistent with other code in Gatsby Interface: 
```suggestion
import * as React from "react"
```

src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,75 @@
...
+  css?: any
+}
+
+export const PlaceholderBox: React.FC<PlaceholderBoxProps> = ({
+  height = "14px",
+  width = "100%",
+  radii = 2,
+  ...props
+}) => {
+  const { animation } = useContext(PlaceholderContext)
```
Elanhant, line 58:
```suggestion
  const { animation } = React.useContext(PlaceholderContext)
```

src/components/Placeholder/index.ts
```diff
@@ -0,0 +1,9 @@
+import {
+  PlaceholderProps as PlaceholderPropsDefinition,
+  PlaceholderBoxProps as PlaceholderBoxPropsDefinition,
+} from "./Placeholder"
+
+export type PlaceholderProps = PlaceholderPropsDefinition
+export type PlaceholderBoxProps = PlaceholderBoxPropsDefinition
+
+export { Placeholder, PlaceholderBox } from "./Placeholder"
```
Elanhant, line 9: This entire file can be just one line 🙂 
```typescript
export * from "./Placeholder"
```

mfrachet, line 9:
```suggestion
export * from "./Placeholder"
```

fk (approved):
<3

Elanhant (approved):

mfrachet:
I got some A11y errors on some other components than this one 🤔 is that the actual state of interface or should I fix them?

Elanhant:
> I got some A11y errors on some other components than this one 🤔 is that the actual state of interface or should I fix them?

Form errors/hints all of a sudden started to fail this a11y check, even though axe, WAVE and Storybook's "Accessibility" tab do not report any issues. I will need to look into that, but you can go ahead!

PR #361, opened 2020-07-22
feat: AreaWithHeading component

Elanhant:
Inspired by [this blog post](https://medium.com/@Heydon/managing-heading-levels-in-design-systems-18be9a746fa3), this PR introduces a new component, `<AreaWithHeading>`, which ideally should save us from having to manually check and fix skipped heading levels. 

`<AreaWithHeading>` keeps track of the current heading level, and as long as you wrap your headings and related content in this component, `<Heading>` will render the appropriate HTML tag.


src/components/AreaWithHeading/AreaWithHeading.tsx
```diff
@@ -0,0 +1,26 @@
...
+export function useHeadingLevel(): number {
+  return React.useContext(HeadingLevelContext)
+}
+
+export type AreaWithHeadingProps = {
+  children?: React.ReactNode
+}
+
+export function AreaWithHeading({ children }: AreaWithHeadingProps) {
+  const parentHeadingLevel = useHeadingLevel()
```
mfrachet, line 18: How can you access the `HeadingLevelContext` since the component is not wrapped in a Context.Provider? 🤔 . I would expect React to throw for the top level component

mfrachet: Yes but where does the top level `AreaWithHeading ` gets the information for the first time since nobody has given it? 🤔

Elanhant: This is what the parameter passed to `React.createContext` is for — it provides the default value in case a consuming component is rendered outside of the context provider :)

https://reactjs.org/docs/context.html#reactcreatecontext

amberleyromo:
So this requires the wrapping of `<AreaWithHeader>` around each `<Heading>` instance as well as as a nested structure (like a `<section>`?

Unless I'm missing something, I think we're just trading complexity here, and I'm extremely wary of this approach. We've gone from having to be consciously mindful of headings (which have a clearly understandable API, essentially, h1 - h2) to abstracting that away and having to be mindful of keeping each heading element in a separate structure? And not being able to see in the code directly what you're building?

Imo I see this causing way more problems than good. I think it's dangerous territory when you start getting into abstracting away accessibility concerns for developer convenience. Whenever I'm looking at a page, I have to think about what heading structure makes sense (it's not always cut and dry, or algorithmically definable). And with this pattern, in making those decisions, we'd have to rely on an implicit hierarchical relationship. I just think that would be so much harder.

amberleyromo:
I'm really interested in what @madalynrose and @marcysutton think about this?

src/components/Heading/Heading.stories.tsx
```diff
@@ -64,3 +65,42 @@ export const Variants = () =>
 Variants.story = {
   decorators: [withVariationsContainer],
 }
+
+export const AutomaticHeadingLevels = () => (
```
madalynrose, line 69: If this is our example of usage for this component, is this what we're advocating actual code to look like? 

If so, I find this tremendously confusing. I can't tell what any of the heading levels actually are by looking at the JSX and I'm concerned we're creating components that will really easily enable unintentional misuse of headings and obscure meaning within the code.

madalynrose:
This is one of the instances where I think the solution is clever and may save _some_ devs time but will increase time spent in code review phase because you can't tell what heading level anything is without firing up the page in a browser. I think drawbacks outweigh the pros of "magic" in this case.

marcysutton:
Circling back to comment here–I agree with @amberleyromo and @madalynrose's sentiments: it seems like this introduces a ton of complexity to avoid dealing with heading levels. The API adding two new components increases the cognitive load quite a bit when h1-h6 headings have a well-known and straight-forward API, and I worry this will make the codebase increasingly difficult to understand. 

If this did move forward, where would it be documented for engineers to know how to use it? I don't think Storybook is an adequate replacement for documentation, but even documenting it somewhere that everyone knows to look may not save it from being overcomplicated. Some more concrete use cases for how this would improve things would help, especially if these downsides are weighed with them (How much time would it save? Whose time would it save? What would be the downside of NOT merging this?).

PR #363, opened 2020-07-25
feat(Modal): changes default size for ModalPanel

Elanhant:
Changes default size for `ModalPanel` to 432px or 100% if screen width is less than that value. I don't think we've been using `maxWidth` anywhere in Cloud, but to be safe I've decided to keep it for now and add a deprecation message.

I have not changed the purple color of the overlay since its shared between normal modals and panels, will take care of that in a different PR later.

Preview available in [Chromatic](https://5d4c6885a869c600201d3c8b-pwghngfmjx.chromatic.com/?path=/story/modal-styledpanel--usage-example).

### DIFF
--- a/src/components/Modal/ModalPanel.tsx
+++ b/src/components/Modal/ModalPanel.tsx
@@ -4,6 +4,7 @@ import React from "react"
 import { keyframes } from "@emotion/core"
 import { ModalContent, ModalContentProps } from "./ModalContent"
 import { ThemeCss } from "../../theme"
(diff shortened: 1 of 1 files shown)


fk:
Just noticed that horizontal scrollbar during "panel opening": 
image

Elanhant:
> Can we (or are we, in a different spot, or is this a Storybook thing) do anything to prevent that?

This is probably related to the way we animate these panels, will look into it when I have free time (which seems to be not too soon unfortunately 😞)

PR #366, opened 2020-07-27
Support icon buttons

Elanhant:
Adds three new components — `IconButton`, `IconAnchorButton` and `IconLinkButton` — counterparts to `Button`, `AnchorButton` and `LinkButton` respectively. I didn't want to overload the existing components with new props and the related complexity but instead introduce three new components. 

The new components accepts most of the "visual" props supported by `Button` (`size` — ~without `"XL"`~, `tone`, `variant`) with the exception of `rightIcon`, `leftIcon`, `loading`, `textVariant`. 

The difference from `Button` is that `IconButton` **requires** `icon` AND `children` props to be passed — the first one is obvious (we do want an "icon" button after all), and `children` are used as an accessible, visually hidden label. 

**NOTE**: the button sizes we currently have in gatsby-interface are not the same as the ones in the design, so the new components are in line with the existing ones and not the design. 

Screen Shot 2020-07-26 at 10 20 30 PM


src/components/Button/Button.stories.tsx
```diff
@@ -55,18 +67,41 @@ export default {
...
+    <NotAvailableForIconButton
+      css={(theme: Theme) => ({ marginTop: theme.space[4] })}
+    />
   </React.Fragment>
 )
 
+function NotAvailableForIconButton({ className }: { className?: string }) {
+  return (
+    <Badge tone="NEUTRAL" className={className}>
+      Not availbale for IconButton
```
fk, line 100: 👍 Kudos for sticking to my crappy spec, Aleksei. 🙏 Don't hate me for it, but we should IMO support the "icon only" variant for that size, too — i don't  I just was too stupid to include it in that _crappy_ spec.

If you want to include it here, icon size would be `32px`.
That said, we don't currently need it.

fk, line 100:
```suggestion
      Not available for IconButton
```

PR #367, opened 2020-07-29
Add plan feature icons

amberleyromo:
- Features
- Collaboration
- Login
- Providers
- Deploy

~@Elanhant I haven't dug into it yet, but do you know off the top of your head why this would happen? Most of these icons I've added aren't framed correctly (see below, cut off on the bottom and right).~

CleanShot 2020-07-29 at 09 15 36

CleanShot 2020-07-29 at 09 15 28

CleanShot 2020-07-29 at 09 15 19

CleanShot 2020-07-29 at 09 15 09

CleanShot 2020-07-29 at 09 14 57


Elanhant:
> @Elanhant I haven't dug into it yet, but do you know off the top of your head why this would happen? Most of these icons I've added aren't framed correctly (see below, cut off on the bottom and right).

I have a guess — the original icons from the designs are probably not 24x24 which makes the SVG that you're using not compatible with `IconSkeleton`. You can ask for the properly sized icons in Figma comments.

amberleyromo:
Of course. They're 32x32. Thanks!

PR #369, opened 2020-07-31
Use "body" font family for <Text />

mxstbr:
The Text component hardcodes the use of the "system" font stack. This unfortunately means that to globally use a different font family we have to override it in the theme like so:

```JS
const theme = {
  ...baseTheme,
  fonts: {
    ...baseTheme.fonts,
    system: '...other fonts...'
  }
}
```

But then manually using `fontFamily: 'system'` no longer shows system fonts. That is really confusing! :confused:

Instead, this PR changes the Text component to use the "body" font stack, which is aliased to "system" by default anyway, so no behavior changes. We can then adjust that font stack without anything being confusing! :tada:

```JS
const theme = {
  ...baseTheme,
  fonts: {
    ...baseTheme.fonts,
    body: '...other fonts...'
  }
}
```


fk:
❤️

PR #371, opened 2020-08-01
feat: SplitButton component

Elanhant:
Adds a new component, `SplitButton`, which is basically just two separate buttons put close together, with the first one being the "primary" action and the second one, "secondary" toggling a dropdown of additional actions. Under the hood it uses `Button`, `useIconButton` from `IconButton` and `DropdownMenu`.

It accepts most of the props supported by `Button` (see `SplitButtonProps` definition), but the primary button must be passed via `buttonLabel` prop, while the `children` should contain a combination of `DropdownMenuItem` and `DropdownMenuLink`. Also required is `dropdownButtonLabel` prop, which should provide an accessible label for the secondary button (dropdown toggle).

@fk @nachiketkumar I have disabled "GHOST" variant for this split button, as I'm not sure if it makes sense for this pattern. Should be easy to bring it back if needed though. We should also probably add some "active" styles for the dropdown toggle (and for buttons in general, actually 🙂).  

#### Basic
Screen Shot 2020-07-31 at 11 21 25 PM

#### Variants
Screen Shot 2020-07-31 at 11 21 31 PM

#### Supports new button "width" prop
Screen Shot 2020-07-31 at 11 21 59 PM

More examples in Storybook: https://deploy-preview-371--gatsby-interface.netlify.app/?path=/story/splitbutton--basic

### Trying out in other projects
You can install pre-release version of this PR by running
```bash
yarn add gatsby-interface@Elanhant-split-button
```
in your project's direcotry.


fk:
🥳😻⚡️

fk:
> We should also probably add some "active" styles for the dropdown toggle (and for buttons in general, actually 🙂).

👍

Elanhant:
> I'm not 100% convinced if we should allow the component's background color here instead of a border…

Good point, haven't thought of that! Going to use a linear gradient hack to overlay the tone background with the white-ish color.

nachiketkumar:
@Elanhant @fk will these continue to support the various button sizes? In particular I'll be using these for the Desktop PoC in the `S` size so I wanted to confirm.

Elanhant:
> @Elanhant @fk will these continue to support the various button sizes? In particular I'll be using these for the Desktop PoC in the `S` size so I wanted to confirm.

Yep! Here's an example: https://deploy-preview-371--gatsby-interface.netlify.app/?path=/story/splitbutton--sizes

It's worth noting, however, that currently `Button` in gatsby-interface has slightly different sizes than the ones in Figma.

PR #372, opened 2020-08-05
Animated InProgressIcon

Elanhant:
Makes `InProgressIcon` animated similar to what we do in Gatsby Cloud for build status icon. 

Like most other icons, this one can be colored by applying some `color` CSS, e.g.
```typescript
import { InProgressIcon, ThemeCss } from "gatsby-interface"

const purpleIconCss: ThemeCss = theme => ({
  color: theme.colors.gatsby,
})

function PurpleInProgressIcon() {
  return <InProgressIcon css={purpleIconCss} />
}
```

### Preview
https://5f2b05a6b9c58d00084532bb--gatsby-interface.netlify.app/?path=/story/icons-single-icons--inprogressicon


fk:
cc @ascorbic

PR #373, opened 2020-08-06
Basic docs on using and adding icons

Elanhant:
This is a very short and basic doc on using and adding icons to gatsby-interface.

### DIFF
--- a/docs/icons.md
+++ b/docs/icons.md
@@ -4,10 +4,29 @@ gatsby-interface mostly relies on two sets of icons:
 * Material Design icons (imported from [`react-icons`](https://github.com/react-icons/react-icons))
 * Gatsby icons (can be found in [/src/components/icons]())
 
-This doc mostly concerns the latter kind of icons.
+This doc mostly concerns the latter kind of icons, specific to Gatsby ecosystem.
 
 ## Icon Sizes
-TODO
+Gatsby icons support a set of preset icons in addition to inheriting based on current font size:
+
+```typescript
+export type IconSize =
+  | "inherit" // 1em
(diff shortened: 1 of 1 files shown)


docs/icons.md
```diff
@@ -0,0 +1,109 @@
+# Icons in gatsby-interface
+
+gatsby-interface mostly relies on two sets of icons:
+* Material Design icons (imported from [`react-icons`](https://github.com/react-icons/react-icons))
+* Gatsby icons (can be found in [/src/components/icons]())
+
+This doc mostly concerns the latter kind of icons, specific to Gatsby ecosystem.
+
+## Icon Sizes
+Gatsby icons support a set of preset icons in addition to inheriting based on current font size:
```
mfrachet, line 10: Not breaking a line is not problematic in this MD generation?

Also I would have written "Gatsby icons supportS" but not sure about that one 🤔

Elanhant: > Not breaking a line is not problematic in this MD generation? Like inlining the content instead of returning to the next line?

Not sure what you mean here 😅 

> Gatsby icons supportS
"icons" is plural, hence "support" without "s" suffix :)

mfrachet: Oh you're right! Thanks for the correction :)

Concerning the first comment, I remember that if you don't skip a line between title and the content in some situations (like github I think), the text is displayed on the same line. I was wondering if it would be the same in the case of storybook :)

Elanhant: Ah I see. Never experienced such problems (except when quoting something using "> some text"), but I'll add lione breaks just in case!

PR #374, opened 2020-08-07
Setup a11y checks in CI

Elanhant:
Adds a new CircleCI check, `a11y-checks`, which uses Storybook's [addon-storyshots-puppeteer](https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer) to check stories for basic a11y issues such as color contrast or missing input labels. This check essentially is the same as "Accessibility" tab in Storybook, but allows us to detect problems proactively rather than discovering them in Gatsby Cloud.


src/components/Combobox/Combobox.stories.tsx
```diff
@@ -119,7 +120,7 @@ export const WithToggleButton = () => {
       <Combobox>
         <ComboboxInput
           aria-labelledby="demo"
-          ref={element => element && element.focus()}
+          ref={element => !isA11yTest() && element && element.focus()}
```
mfrachet, line 123: I'm not sure to understand why do we need to add code related to accessibility inside the stories?

Elanhant: `test:a11y` command uses a Storybook addon to run Puppeteer + axe checks on each Storybook story; the only alternative would be to have a set of separate, dedicated pages for a11y tests which I think is an overkill for now, especially since Storybook already supports a11y checks on each story ("Accessibility" tab in addons panel).

fk:
<3

PR #376, opened 2020-08-11
feat: support 'size' prop for some form elements

Elanhant:
POC for supporting different sizes for some of the form elements.

This PR adds support for a new `size` prop to the following components:
* `StyledInput` (and `InputFieldBlock`)
* `StyledSelect` (and `SelectFieldBlock`)
* `StyledTextArea` (and `TextAreaFieldBlock`)
* `ComboboxInput`

The supported values mirror those of `labelSize`: `S`, `M` and `L`. In terms of height and font size those correspond to `ButtonSize` of `M`, `L` and `XL` respectively (supporting an equivalent to button's `S` size doesn't seem viable to me, as the font size is going to be too small in that case). 

### Open questions:
* Could we omit current `labelSize` altogether and use `size` instead, coupling label and control sizes? For now I have made `labelSize` fall back to `size` if no value has been passed.
* What about other form components, such as single checkbox field or checkbox group or radio button? Should we support different sizes for those and how should they look? This can be done in a separate PR.

#### Input
Screen Shot 2020-08-11 at 1 16 31 PM

#### Select
Screen Shot 2020-08-11 at 1 16 37 PM

#### Text area
Screen Shot 2020-08-11 at 1 16 49 PM

#### Combobox input
Screen Shot 2020-08-11 at 1 15 45 PM

### DIFF
--- a/src/components/Combobox/Combobox.stories.tsx
+++ b/src/components/Combobox/Combobox.stories.tsx
@@ -10,8 +10,13 @@ import {
 } from "./"
 import { Theme } from "../../theme"
 import { ComboboxOptionProps } from "./Combobox"
-import { boolean, text } from "@storybook/addon-knobs"
-import { isA11yTest } from "../../utils/storybook"
+import { boolean, text, radios } from "@storybook/addon-knobs"
+import {
+  isA11yTest,
+  radioKnobOptions,
+  withVariationsContainer,
+} from "../../utils/storybook"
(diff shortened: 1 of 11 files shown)


fk:
<3 Most probably won't be able to review this today/night …
<3!!!

Elanhant:
> <3 Most probably won't be able to review this today/night …

No worries! Rapid PRs don't always need rapid reviews 😉

PR #377, opened 2020-08-13
Adding campaign icon

mfrachet:
Adding the Campaign Icon:

Screenshot 2020-08-13 at 10 26 12


__tests__/__snapshots__/icons.test.js.snap
```diff
@@ -212,6 +212,28 @@ exports[`icons <BusinessPlanIcon /> renders unchanged 1`] = `
 </div>
 `;
 
+exports[`icons <CampaignIcon /> renders unchanged 1`] = `
```
Elanhant, line 215: We really need to get rid of these snapshots, Chromatic is much better 🙂

PR #378, opened 2020-08-13
Neutral background for modals

mfrachet:
Screenshot 2020-08-13 at 10 41 15

### DIFF
--- a/src/components/Modal/StyledModal.stories.tsx
+++ b/src/components/Modal/StyledModal.stories.tsx
@@ -15,9 +15,12 @@ import {
   StyledModalBody,
   StyledModalActions,
   StyledModalVariant,
+  StyledModalHeaderLayout,
 } from "./"
 import { Theme } from "../../theme"
 import { Button } from "../Button"
+import { Heading } from "../Heading"
+import { Text } from "../Text"
 
 export default {
   title: `Modal/StyledModal`,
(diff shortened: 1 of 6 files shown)


Elanhant:
Is there any particular case where we want this? You can DM in Slack if there's a Figma link or something 🙂

Elanhant:
Never mind, I think I know what this is for. Will comment in Figma.

Elanhant:
Actually no, no need to comment in Figma (at least for now). I think what you need can be achieved by just passing a custom element as `children` to `<StyledModalHeader>`. You can do all the styling in that custom element.

mfrachet:
Yes but it will break the semantic title I think nope?

Elanhant:
> Yes but it will break the semantic title I think nope?

Not if you use a `<span>` or a `<div>` 🙂 
```jsx
<StyledModalHeader>
  <span css={textCss}>
    <MdVisibility css={iconCss} />
    New feature announcement
  </span>
</StyledModalHeader>
```

mfrachet:
Yeah but the header is already a heading: https://github.com/gatsby-inc/gatsby-interface/blob/main/src/components/Modal/StyledModal.tsx#L58 and heading by default is a H2 https://github.com/gatsby-inc/gatsby-interface/blob/main/src/components/Heading/Heading.tsx#L39 ?

Elanhant:
> Yeah but the header is already a heading

Are you planning to put the actual feature title (e.g. "Introducing Gatsby Cloud Workspaces") into `StyledModalHeader`? No need to do that, it should be a part of the modal body. Think of it like this:
* "New feature announcement" is the _modal header_, it describes what the entire modal is about
* Feature title is, well, _the feature title_ — it describes what is the actual feature; it is semantically coupled with the actual content of the modal

mfrachet:
Oh I see! We can approach it that way 😊 Thank you! Tomorrow, I'll revert these changes to only keep the modal overlay in the background! Thanks for your help

src/components/Modal/StyledModal.stories.tsx
```diff
@@ -31,14 +31,20 @@ export default {
...
   decorators: [
-    disableAnimationsDecorator,
     story => (
       <React.Fragment>
         <Global
           styles={(theme: Theme) => [
             {
               body: { background: theme.colors.grey[20] },
             },
+            isChromatic() && {
```
Elanhant, line 41: `disableAnimationsDecorator` is doing exactly the same thing, why remove it? 🙂

mfrachet: Mmh, this have been introduced by the rebase I think 🤔 . I didn't moved these part of the file. Going to re-rebase on master

mfrachet: oooh, rebasing master, not main, I forgot about that 😓

PR #379, opened 2020-08-25
feat: BuildLogsList component

Elanhant:
The main goal of this PR is to add a component to display structured logs similar to what we already have in Gatsby Cloud:
Screen Shot 2020-08-25 at 3 30 12 PM

The new component is called `BuildLogsList` and requires just two props for now:
* `aria-label` — an accessible label for the logs list (e.g. "Build activities")
* `logItems` — an array of objects describing each log entry

### Log items
Each log entry should have a structure similar to structured logs from Gatsby Cloud (minus some fields not used at the moment):
``` typescript
// Represents the most basic log entry
export type BuildLogItem = {
  id: string
  message?: string | null
  level?: StructuredLogLevel | null
  // More detailed log for activities such as creating image thumbnails
  activity?: BuildActivity | null
  filePath?: string | null
  context?: { [k: string]: string } | null
}

export type BuildActivity = {
  id: string
  name?: string | null
  message?: string | null
  type: BuildActivityType
  status: ActivityStatus
  duration?: number | null
  current?: number | null
  total?: number | null
}
```
For enums like `StructuredLogLevel` see `src/components/BuildLogs/types.ts`


src/components/BuildLogs/BuildLogsList.tsx
```diff
@@ -0,0 +1,44 @@
...
+            {logItem.activity ? (
+              <BuildActivityEntry {...logItem} activity={logItem.activity} />
+            ) : (
+              <BuildStandardLogEntry
+                level={logItem.level}
+                message={logItem.message}
+                context={logItem.context}
+                filePath={logItem.filePath}
+              />
+            )}
```
mfrachet, line 38: I'm wondering if we couldn't pass `children` instead and let the host application show the lines it wants? This would allow to have a component that is not bound to any data shape. I think it can be useful since interface is an external dependency that can be used in different projects

src/components/BuildLogs/BuildStandardLogEntry.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const baseCss: ThemeCss = theme => ({
+  [theme.mediaQueries.tablet]: {
+    paddingRight: theme.space[10],
+  },
+})
+
+const messageCss: ThemeCss = theme => ({
+  // Necessary to set min-width to 0 so that the message does not overflow the row
+  minWidth: 0,
```
mfrachet, line 17: This trick is always weird to me. Make me think of text ellipsis on display flex 😅

mfrachet (approved):
Looks okay to me, leaving @fk checking if the styles are consistent 😊 

Also, I'm wondering if we could imagine a specific accessible keyboard experience for this kind of lists?

src/components/BuildLogs/BuildLogsList.tsx
Elanhant: Currently, this components has a very specific use case — displaying logs for a Gatsby site, and therefore I'm hesitant to making any early abstractions until we actually try this out in the wild 🙂

Elanhant:
> Also, I'm wondering if we could imagine a specific accessible keyboard experience for this kind of lists?

Seems like adding a `role="log"` is the only real requirement (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_log_role), no specific keyboard requirements. I will add that.

fk:
> Looks okay to me, leaving @fk checking if the styles are consistent 😊

LGTM, too!
Ty for the ping, @mfrachet! 🤗

PR #382, opened 2020-09-02
Add visually hidden component, export a11y styles

amberleyromo:
- Add a visually hidden component to the library.
  - This applies the `visuallyHiddenCss` style to a `<span>` and wraps children. This is a common and necessary pattern, it makes sense to me to have this defined in our component library. Interested to hear thoughts!
- Export a11y styles (`visuallyHiddenCss` and `resetvisuallyHiddenCss`).
  - I don't really see a reason not to keep this here, and reference wherever needed in docs/dashboard/any of the other properties. (Occurred to me in the splitting of static and dashboard code, cc @mfrachet)


fk (approved):
Yay! Happy to help replacing https://github.com/gatsby-inc/mansion/blob/1ab41a3bc5e1f1fdd96e0b51f506e5326c6b82b5/cloud/gatsbyjs.com/src/utils/styles.js#L268-L279!

fk:
Wondering if I'll still be able to do this: https://github.com/gatsby-inc/mansion/blob/1ab41a3bc5e1f1fdd96e0b51f506e5326c6b82b5/cloud/gatsbyjs.com/src/components/blog/meta.js#L42 ?

amberleyromo:
Yep, this PR is also exporting the `visuallyHiddenCss` from `gatsby-interface` (The same styles living in multiple places).

- interface: https://github.com/gatsby-inc/gatsby-interface/blob/main/src/stylesheets/a11y.ts#L3
- mansion: https://github.com/gatsby-inc/mansion/blob/1ab41a3bc5e1f1fdd96e0b51f506e5326c6b82b5/cloud/gatsbyjs.com/src/utils/styles.js#L268

So in the case you linked, you would import that style from gatsby-interface instead: https://github.com/gatsby-inc/mansion/blob/1ab41a3bc5e1f1fdd96e0b51f506e5326c6b82b5/cloud/gatsbyjs.com/src/components/blog/meta.js#L3

src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,13 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core"
+import { visuallyHiddenCss } from "../../stylesheets/a11y"
+
+export type VisuallyHiddenProps = {
```
Elanhant, line 5: We should allow any props supported by `span` for cases where we want to pass an `id` to link the contained text to some other element via `aria-describedby` or `aria-labelledby` (we do this in several places in Cloud).
```typescript
export type VisuallyHiddenProps = React.ComponentPropsWithoutRef<"span">
```

src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,13 @@
...
+  children: React.ReactNode
+}
+
+function VisuallyHidden({ children }: VisuallyHiddenProps) {
+  return <span css={visuallyHiddenCss}>{children}</span>
+}
+
+export default VisuallyHidden
```
Elanhant, line 13: We should use named exports instead, makes it easier when it comes to searching for a component or re-exporting via index files.

src/components/VisuallyHidden/index.ts
```diff
@@ -0,0 +1,4 @@
+import { VisuallyHiddenProps as VisuallyHiddenPropsDefinition } from "./VisuallyHidden"
```
Elanhant, line 1: Once `VisuallyHidden` is exposed as a named export, we can use a wildcard export here:
```typescript
export * from "./VisuallyHidden"
```

src/components/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,39 @@
+/** @jsx jsx */
```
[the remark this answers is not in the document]
amberleyromo: My bad, I should have noticed! Maybe we should touch on this in our next FE catchup meeting, for everyone?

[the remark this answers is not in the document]
Elanhant: Great idea! I will add it to the agenda. Hopefully I will also get to work on improving docs during the upcoming sprint.

fk:
> So in the case you linked, you would import that style from gatsby-interface instead

Oh nice, thanks for making old man see! 🤗 
(I _was_ looking for that export when I left my review :D 🤦‍♂️)

PR #383, opened 2020-09-02
fix: small fixes in StyledCheckbox, StyledRadio and RawLogs

Elanhant:
A couple of tiny fixes and improvements:
* Fix typings for `RawLogs`, allowing React nodes as log item's `message` property 
* Force `box-sizing: border-box` on `StyledCheckbox` and `StyledRadio` in case this is not a default style (this is the case with Desktop)
* Add a new story with `StyledModal` usage example
* Remove stories for deprecated form components (`*Field` compound components such as `InputField` or `SelectField`)
* `BuildLogsList` now should correctly parse URLs in parsed list items


fk:
Sorry for not giving this a proper review, @Elanhant! 🙏

PR #386, opened 2020-09-15
Fixing builds logs list

mfrachet:
## Inside the PR

- Add markdown to the messages we get to highlight some of the content
- Line break on the message on the first ":" to create something that looks like a title
- Line break on the first line (splitted by "\n) owning a "." as its last character
- Compute code blocks based on line number "1 |" and " |" strings

## Note

I'm a 0 dealing with regex and I'm 80% unsure about what I did 😬 - my biggest problem being at the (input) string consistency level.

## Screenshot

build-log-list


src/components/BuildLogs/__tests__/FormattedMessage.spec.tsx
```diff
@@ -0,0 +1,140 @@
+import React from "react"
```
mfrachet, line 1: ℹ️ Snapshotting the content to make sure what we get looks coherent

src/components/BuildLogs/FormattedMessage.tsx
```diff
@@ -0,0 +1,68 @@
...
+// Check for "1 |" presence to create a code block
+// Stores every last index owning a " |" string
+// to close the code block at the good position
+// IMPORTANT: mutates the stringPartsByLines array
+const formatCodeBlocks = (stringPartsByLines: string[]) => {
+  let lastIndexFound = -1
+  for (let i = 0; i < stringPartsByLines.length; i++) {
+    const str = stringPartsByLines[i]
+
+    if (str.includes(`1 |`)) {
```
Elanhant, line 15: Hm, I think we need a regex for this — we could potentially get a message with line numbers starting somewhere else. 
```suggestion
    if (str.matches(/(\s|\t)*\d+\s\|/)) {
```

src/components/BuildLogs/FormattedMessage.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react"
+import Markdown from "markdown-to-jsx"
+
+const isLastChar = (char: string, str: string) => str[str.length - 1] === char
+
+// Check for "1 |" presence to create a code block
+// Stores every last index owning a " |" string
+// to close the code block at the good position
+// IMPORTANT: mutates the stringPartsByLines array
```
Elanhant, line 9: Can we do this without mutating `stringPartsByLines`?

src/components/BuildLogs/FormattedMessage.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <p>
+        <Markdown>{message}</Markdown>
+      </p>
+    )
+  }
+
+  const firstLine = stringPartsByLines[0]
+
+  // Line breaking to create a title at the first ":" met
+  if (firstLine.includes(":")) {
```
Elanhant, line 54: What do you think about using `.endsWith` instead? I am a bit concerned that the first line might also include a colon somewhere in between.

src/components/BuildLogs/FormattedMessage.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return <Markdown>{message.replace(`:`, `:\n\n`)}</Markdown>
+  }
+
+  // Line breaking to create a title when the first line ends with "."
+  if (isLastChar(".", firstLine)) {
```
Elanhant, line 59:
```suggestion
  if (firstLine.endsWith('.')) {
```

Elanhant (approved):
LGTM!
I have one nit, but we can handle it in a follow-up PR — the "code" parts of messages (those wrapped in "\`" and "```") look too small compared to the surrounding text, perhaps we should set a bigger font size for `<code>` tags?

mfrachet:
@Elanhant fixing the size:

Screenshot 2020-09-16 at 14 53 26

PR #388, opened 2020-09-17
feat(DropdownMenu): support new 'size' prop

Elanhant:
Adds a new prop, `size`, for `DropdownMenuItems` and `DropdownMenuItemsLowLevel` components that allows to control the width of dropdown menu list.

Supported values are: 
* `AUTO` — width is defined by menu items width
* `MAX_CONTENT` — sets menu's width to `max-content`
* `S` — sets menu's width to `12rem`
* `M` — sets menu's width to `15rem`
* `L` — sets menu's width to `20rem`

Default value is `L` (it corresponds to the current width of our dropdown menus).

**NOTE:** `S`, `M` and `L` behave as `MAX_CONTENT` on screen widths smaller than `phablet` (`550px`).

### Images
localhost_6006_iframe html_id=dropdownmenu--sizes (1)


fk (approved):

greglobinski (approved):
LGTM, 

although I'm not sure if we need these 'artificial' variants, I mean the S, M, L set outside of the real app context, variants at all. Maybe an approach where the `width` is always based on `max-content` but never wider than `max-width` or narrower than `min-width`?

Elanhant:
> although I'm not sure if we need these 'artificial' variants, I mean the S, M, L sizes set outside the real app context at all

This is a great observation, and I had a similar concern. I will see if we can just try and always base the width on `max-content`. Perhaps we can omit `S` and `M` and keep `L` for backwards compatibility for now (but rename it):
* `AUTO` — width is defined by menu items width
* `MAX_CONTENT` — sets menu's width to `max-content`
* `LEGACY` — sets menu's width to `20rem` (corresponds to the current width of our dropdown menus)

Then we can start adding `size="MAX_CONTENT"` to dropdowns in Cloud and see if they still look OK. Once we are sure that there aren't any issues we can make `MAX_CONTENT` the default value.

PR #392, opened 2020-09-28
Upgrade to Storybook 6

Elanhant:
This PR updates Storybook-related dependencies to stable v6.

There shouldn't be any breaking changes here — most of this PR is just adopting our existing stories to the [suggested format](https://storybook.js.org/docs/react/writing-stories/introduction#using-args) with a template and args.
  
**NOTE**: seems like something is wrong with our a11y CI check — it is failing for form field stories that demonstrate error message; those messages are animated using CSS, and we do disable those animations (or rather make them instantaneous), but the check seem to have started ignoring that. I will need to look into this deeper, but most probably as a separate PR.


fk (approved):
🎉

mfrachet (approved):

PR #393, opened 2020-09-28
feat(dropdown): Add `DropdownHeader`, `DropdownDivider`

fk:
* add `DropdownHeader`, renders as `h6`
* add `DropdownDivider`, renders as `div`
* add the two to the `Sandbox` story
* capture a couple `@TODO`s in source

image

Along the way:

* add vertical padding to the dropdown container — this aligns the first and last action items' visual size with the rest of the dropdown items
* change "enter" animation to be snappier:
  * remove rotation
  * animate opacity and scale instead
  * use `theme` value for `transition.speed`
  * note: let's add an "exit" animation, too (@mfrachet) 
* switch story example dropdown item labels ("First, Second, …" -> "Item 1, Item 2, …")
* simplify `Basic`, `StyledButton` and `WithComponentPlaceholder` stories
* link Figma design


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -115,3 +117,13 @@ export function DropdownMenuPopover(props: DropdownMenuPopoverProps) {
 export function useDropdownMenuContext() {
   return useMenuButtonContext()
 }
+
+export function DropdownDivider(props) {
+  return <div {...props} css={dropdownDividerCss} />
+}
+
+export const DropdownHeader = ({ children, ...props }) => (
+  <h6 {...props} css={dropdownHeaderCss}>
```
gillkyle, line 126: Would it be helpful to have the tag that renders be prop driven with h6 as the default? Maybe that's an unnecessary thing and I don't know all the accessibility implications either. But for instance, I can imagine a situation where a page would have only an h1, h2, and h3 and this in the header would skip straight to h6. 🤷

fk:
@mfrachet @Elanhant May I ask for assistance with the TypeScript Kung Foooooo? 🙏 
Also unsure about forwarding props for the two new components.

gillkyle (commented):
Going to go do some reading on headings and what tags the menus in gatsby-interface render now, but had 1 question is all!

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -115,3 +117,13 @@ export function DropdownMenuPopover(props: DropdownMenuPopoverProps) {
 export function useDropdownMenuContext() {
   return useMenuButtonContext()
 }
+
+export function DropdownDivider(props) {
+  return <div {...props} css={dropdownDividerCss} />
```
Elanhant, line 122: We should use `<hr>` element as well as [`role="separator"`](https://www.digitala11y.com/separator-role/)
```suggestion
  return <hr role="separator" {...props} css={dropdownDividerCss} />
```
Then we can provide a type for its props using a generic type from React:
```typescript
export type DropdownDividerProps  = React.ComponentPropsWithoutRef<"hr">
```
This will let TS know that `DropdownDivider` accepts all the props/attributes supported by `<hr>`.

src/components/DropdownMenu/DropdownMenu.tsx
fk: Great call, Kyle. 🙏 
I wondered about that when I wrote that, then forgot about it.
Probably shouldn't be an `h*` at all? I'm feeling old. ;-)

src/components/DropdownMenu/DropdownMenu.tsx
fk: 😬 That's what I had here, first. Removed it because I wondered if this _fully_ qualifies as a "thematic break" — from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr:

> The HTML <hr> element represents a thematic break between paragraph-level elements: for example, a change of scene in a story, or a shift of topic within a section.

I didn't consult the actual spec for `hr` yet.

fk: Then again, same source, emphasis mine:

> Historically, this has been presented as a horizontal rule or line. While it may still be displayed as a horizontal rule in visual browsers, this element is now defined in semantic terms, rather than presentational terms, **so if you wish to draw a horizontal line, you should do so using appropriate CSS.**

src/components/DropdownMenu/DropdownMenu.tsx
Elanhant: What if we just render our own `Heading` component? It allows to pick which `h*` to render and doesn't apply too many styles that we'd need to override.

gillkyle: @amberleyromo had some good feedback in our frontend chapter meeting. She mentioned it seemed like the idea of this component is to take one of these kinds of menus:

image

And flatten it into one list (where share would be the heading in the screenshot above), meaning that rather than a heading, the element used here would actually be a sublist. I'm going to try and find some other examples.

src/components/DropdownMenu/DropdownMenu.tsx
gillkyle: I asked about this in our frontend chapter meeting as well and the general consensus was that since the `<hr />` should semantically separate content and this list would have separated content, that using the `<hr />` would be a good way to go.

src/components/DropdownMenu/DropdownMenu.tsx
gillkyle: ## Stripe

So Stripe uses `h1`'s very liberally (they have 31 on their homepage), that doesn't seem like the example to follow. 

image

## Segment

Segment uses `span`s which seems like maybe is just sort of a cop out 🤷 

image

gillkyle: I suppose it's worth noting that we use heading elements in our menu on the marketing pages

src/components/DropdownMenu/DropdownMenu.tsx
fk: ^ fixing that now! 🙏

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -115,3 +118,15 @@ export function DropdownMenuPopover(props: DropdownMenuPopoverProps) {
...
   return useMenuButtonContext()
 }
+
+export type DropdownDividerProps = React.ComponentPropsWithoutRef<"hr">
+
+export function DropdownDivider(props) {
+  return <hr {...props} css={dropdownDividerCss} />
+}
+
+export const DropdownHeader = ({ children, ...props }) => (
```
Elanhant, line 128: Can I ask you to provide a props type for `DropdownHeader`, something like this:
```typescript
export type DropdownHeaderProps = HeadingProps

export const DropdownHeader = ({ children, ...props }: DropdownHeaderProps) => (
  <Heading {...props} css={dropdownHeaderCss}>
    {children}
  </Heading>
)
```
I have previously explained my reasoning for this in https://github.com/gatsby-inc/gatsby-interface/pull/275#discussion_r405710902

gillkyle:
Woops, clicked the request review button before actually making any changes 😬  ..gimme a sec..

src/components/DropdownMenu/DropdownMenu.tsx
gillkyle: > I have previously explained my reasoning for this in #275 (comment)

Thanks @Elanhant! That was super helpful to link to that response! Just updated this and the divider component above it as well 🙂

Elanhant (approved):

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -121,11 +121,13 @@ export function useDropdownMenuContext() {
 
 export type DropdownDividerProps = React.ComponentPropsWithoutRef<"hr">
 
-export function DropdownDivider(props) {
+export function DropdownDivider(props: DropdownDividerProps) {
```
fk, line 124: 😬 🤦 🙏

fk:
Just checked preview, we need to get Futura out of there :D — will circle back later ✌️ (but feel free to grab) 

image

fk:
Sweet, thanks everyone! <3

PR #395, opened 2020-09-30
feat(BuildLogs): Display docsUrl and errorUrl for errors

julienp:
Show more information on errors when it's available:

Screenshot 2020-10-01 at 10 24 09

### DIFF
--- a/src/components/BuildLogs/FormattedLogMessage.tsx
+++ b/src/components/BuildLogs/FormattedLogMessage.tsx
@@ -3,10 +3,12 @@ import { jsx } from "@emotion/core"
 import { BuildLogItem, StructuredLogLevel } from "./types"
 import { ThemeCss, Theme } from "../../theme"
 import { FormattedMessage } from "./FormattedMessage"
+import { Link } from "../Link"
+import { Text } from "../Text"
 
 export type FormattedLogMessageProps = Pick<
   BuildLogItem,
-  "level" | "message"
+  "level" | "message" | "docsUrl" | "location" | "filePath" | "errorUrl"
 > & {
   className?: string
 }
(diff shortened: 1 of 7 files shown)


DSchau:
Hot damn this is good stuff 🔥

src/components/BuildLogs/FormattedLogMessage.tsx
```diff
@@ -3,22 +3,41 @@ import { jsx } from "@emotion/core"
...
+  filePath,
+  errorUrl,
   className,
 }: FormattedLogMessageProps) {
   return (
     <div css={getMessageCss(level)} className={className}>
       <FormattedMessage rawMessage={message || ""} />
+      {docsUrl && (
+        <Text>
+          For more details see <Link href={docsUrl}>{docsUrl}</Link>
```
julienp, line 30: This ends being red for some errors because of the `&:first-of-type` rule on FormattedLogMessage. I am not sure what the best way to address this is?

Elanhant: A possible solution would be to extract the `"&:first-of-type"` style into a separate theme function, wrap `<FormattedMessage>` in an additional `<div>` and apply the extracted style to that `<div>`. 
```typescript
// getMessageCss helper

const headlineCss: ThemeCss = theme => ({
  "&:first-of-type": {
    color: getLogLevelColor(level)(theme),
    fontSize: theme.fontSizes[1],
  },
})

const messageCss: ThemeCss = theme => {
  // ...
}

return { messageCss, headlineCss }
```

```jsx
// FormattedLogMessage component

const { messageCss, headlineCss } = getMessageCss(level)

return (
  <div css={messageCss} className={className}>
    <div css={headlineCss}>
      <FormattedMessage rawMessage={message || ""} />
    </div>
  </div>
)
```

julienp: I ran into some conflicts between making the "first p is the title" and "add a margin except on the last element" work together. 

What I ended up with is firstly to ensure the markdown always generates a structure of `div > p` even for single lines, and secondly targeting the title to exactly that and overriding the marginBottom for the single line special case.

mfrachet (approved):
Approving but I don't want to take over @Elanhant on this

Elanhant (approved):

PR #400, opened 2020-10-05
fix: various TypeScript fixes

Elanhant:
Fixes several TypeScript errors that might occur when using Gatsby Interface:
* Using a React element as `title` for `CardHeader` would result in an error, since the prop type from `CardHeaderProps` conflicts with the type for `title` attribute on `<div>`/`<header>` elements. 
* `ComboboxPopover` had incorrect type specified for its `ref` — it was `HTMLInputElement` instead of `HTMLDivElement`.
* `DropdownMenuLink` would not accept any `as` prop other than `"a"`; I had to use ReachUI's utility helper `forwardRefWithAs` in order to fix this.

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -68,7 +68,7 @@ const headerCss: ThemeCss = theme => ({
   lineHeight: 1,
 })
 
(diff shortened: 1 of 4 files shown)


fk (approved):
👍 to fixing fixes ;) <3

PR #401, opened 2020-10-05
fix: replace keyframes with spring animation for DropdownMenu

Elanhant:
While using CSS keyframes to animate an opened dropdown menu is the simplest solution, it breaks our Jest tests that check for dropdown menu items to be visible (using Testing Library matchers). This is due to Jest environment being unable to detect style changes coming from keyframes. 

To fix this and allow us to test for menu visibility again, I have switched to [`useSpring`](https://www.react-spring.io/docs/hooks/use-spring) from React Spring. It explicitly injects styles into DOM which allows to actually check whether the animated menu is visible. 

**NOTE:** I have initially tried to animate with `useTranisition`, but it resulted in some unpleasant bugs when the user quickly reopens the menu after closing (e.g. when using keyboard).

@fk be my guest to play with the spring configuration if you want 🙂 I used the same start/end values that you have specified for the keyframe, but the spring itself probably behaves a bit different from what you envisioned.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,9 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "scripts": {
-    "test": "jest",
+    "test": "jest --env=jest-environment-jsdom-sixteen",
     "test:a11y": "jest __tests__/AccessiblityStoryshots.test.js --testPathIgnorePatterns=[]",
-    "test:watch": "jest --watch",
+    "test:watch": "jest --watch --env=jest-environment-jsdom-sixteen",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "tsc": "tsc",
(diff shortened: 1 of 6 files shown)


fk (approved):
👍 https://gatsby-interface.netlify.app/?path=/story/dropdownmenu--sandbox is a tiny bit snappier though, which I'd love to preserve. I guess we'd need to make the spring a bit springy-er? :P

Elanhant:
> 👍 https://gatsby-interface.netlify.app/?path=/story/dropdownmenu--sandbox is a tiny bit snappier though, which I'd love to preserve. I guess we'd need to make the spring a bit springy-er? :P

@fk what about now? 🙂  https://5d4c6885a869c600201d3c8b-lcjlgbtlbl.chromatic.com/?path=/story/dropdownmenu--basic

fk:
> what about now?

👍 🙏

PR #404, opened 2020-10-06
fix(BuildLogs): Handle multiple code blocks

julienp:
The generated markdown was invalid when there were multiple code blocks.

Before
Screenshot 2020-10-06 at 13 57 39

After
Screenshot 2020-10-06 at 14 00 21


Elanhant (approved):
👍

PR #405, opened 2020-10-07
fix(form): various performance optimizations across form components

Elanhant:
This PR contains several performance improvements for form components, some of them are based on these issues in Formik repo: https://github.com/formium/formik/issues/2268 and https://github.com/formium/formik/issues/1026.

Summary of improvements:
* `useAriaFormField` and `useAriaFormGroupField` now wrap their inner logic in `useCallback` and `useMemo` hooks — the values in objects such as `labelProps`, `hintProps` etc do not change very often, yet every call to the `useAriaForm` results in a new object being created (new `labelProps`, `hintProps` etc), leading to false prop changes. Memoizing these helps to reduce the amount of unnecessary re-renders.
* `CheckboxGroupConnectedField` wraps its inner `onBlur` and `onChange` handlers in `useCallback` for the same reason as the previous item. If we have an `InputConnectedField` in the same form as `CheckboxGroupConnectedField`, each time the user types something in the former, the latter gets re-rendered because `onBlur` and `onChange` are re-created. `useCallback` prevents that.
* `InputFieldBlock` and `TextAreaFieldBlock` now use `useAriaFormField` directly instead of relying on `FormFieldBlock` and its render children — these two fields are the most prone to performance issues on value update, since the user types inside them and each keystroke results in a value change. Skipping render function allows to speed re-renders a little bit.
* `useConnectedField` is now using a custom "fast" wrapper for Formik's `useField` based on comments from one of the issues mentioned in the beginning of this PR.

### DIFF
--- a/src/components/form-skeletons/hooks/useAriaFormField.tsx
+++ b/src/components/form-skeletons/hooks/useAriaFormField.tsx
@@ -1,3 +1,4 @@
+import * as React from "react"
 import { getHintId, getErrorId, getErrorAriaLiveAttribute } from "../utils"
 import { ErrorValidationMode } from "../types"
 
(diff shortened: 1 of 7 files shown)


fk (approved):
✨ LGTM ;)

Elanhant:
@mfrachet @greglobinski could you please take a look at this PR when you have time?

mfrachet (approved):
I really appreciate you take time to improve this, thank you 😊 

One thing that I would find very helpful when improving performances, is to have a before / after screenshots showing maybe the React profiler in action with "we had 30 renders with 10 commits and now we only have 2 renders with only 1 Commit"

Elanhant:
> One thing that I would find very helpful when improving performances, is to have a before / after screenshots showing maybe the React profiler in action with "we had 30 renders with 10 commits and now we only have 2 renders with only 1 Commit"

Good point! 
Using a `<FieldArray>` allows for a better demonstration on what has changed. Here is the screenshot of the story I have tested:
localhost_6006_iframe html_id=form-connected-fields--array-values-example viewMode=story
The story renders three field groups, each consisting of two `InputConnectedField` components. After starting the profiler, I have typed "typingintosecondfield" into the "name" field from the second group. Profiling is stopped when I finish typing.

Here is the render timeline comparison:
### `main` branch:
Screen Shot 2020-10-14 at 8 29 02 AM

### This PR:
Screen Shot 2020-10-14 at 8 29 10 AM
 
And here is the ranked breakdown of renders at the same moment when I type one of the characters:
### `main` branch:
Screen Shot 2020-10-14 at 8 39 14 AM

### This PR:
Screen Shot 2020-10-14 at 8 40 11 AM

Avoiding "render children" from `FormFieldBlock` helps to even out the render spikes when typing.

mfrachet (approved):
Nice improvements, the renders take way less time 😊

PR #408, opened 2020-10-09
feat(Heading): support Inter font

Elanhant:
Disclaimer: I am a noob when it comes to fonts, so might be doing something stupid here 😅 

This PR adds a new font token to our theme, `headingUI`, that uses Inter instead of Futura. This font then is added to our `Heading` component via a new prop, `fontVariant`:
* `"DISPLAY"` — will render the heading using Futura
* `"UI"` — will render the heading using Inter

**NOTE:** consumer applications will have to install Inter on their own, at least for now. I am not exactly sure if we want to distribute it with Gatsby Interface (also need to read through the license).


assets/fonts/inter/stylesheet.css
```diff
@@ -0,0 +1,27 @@
+  @font-face {
+    font-family: 'Inter';
+    font-style:  normal;
+    font-weight: 400;
+    font-display: swap;
+    src: url("./Inter-Regular.woff2?v=3.12") format("woff2"),
```
fk, line 6: Yay, good to see the version is captured! 👍

fk: Seems https://rsms.me/inter/ has 3.15 for download already … :-P

fk:
> NOTE: consumer applications will have to install Inter on their own, at least for now. I am not exactly sure if we want to distribute it with Gatsby Interface (also need to read through the license).

https://github.com/philipbelesky/inter-ui looks like it'd be fine to include it …
Maybe we could also use that package? 🤔

assets/fonts/inter/stylesheet.css
fk: Added a commit updating stylesheet and fonts to latest (v3.15). Thinking about it more, will add another commit adding `SemiBold` and `ExtraBold` — the latter is something we'll need for "public .com" stuff when making the switch away from Futura; `SemiBold` is already in use by a couple components: 
image

fk: 6f6df32

fk:
Argh now I can't approve this anymore ;-)
I think this is good to go; couple of thoughts: 

* Wondering how much sense it makes to investigate the variable font.
* Great to see the "Inter Web" fonts are already subsetted; https://github.com/philipbelesky/inter-ui#latin made me wonder if we should think about dropping some more weight.
* When moving UI components to use Inter instead of Futura via `fontVariant="UI"`, we most likely need to adjust the `fontSize` down one step on the scale — Futura's [small x-height](https://www.researchgate.net/figure/A-comparison-of-x-height-proportions-of-Futura-Times-New-Roman-Verdana-and-Sitka-All_fig2_309754147) being the reason for that — Inter will appear much bigger at the same `fontSize`
* We IMO should revisit variants: image
  * `LIGHT` was implicitly introduced, IIRC an audit shouldn't produce many usages; do we support that for Inter?
  * `EMPHASIZED` and `PRIMARY` IMO are slighly confusing — I personally interpret `PRIMARY` as `h1` … hmm, will think about it more

Elanhant:
> We IMO should revisit variants:

FYI, quick search in .com code gives these results:
* `LIGHT` variant is used 15 times, all but 1 are in code related to static pages and not Gatsby Cloud
* `EMPHASIZED` variant is used 28 times, 3 of them in Gatsby Cloud

Elanhant:
> Wondering how much sense it makes to investigate the variable font.

I have no opinion here — we can investigate it or we can postpone it 🙂 

> Great to see the "Inter Web" fonts are already subsetted; https://github.com/philipbelesky/inter-ui#latin made me wonder if we should think about dropping some more weight.

We can try this in Gatsby Cloud, but probably not on .com static pages — didn't they use to include some translated content? That would probably break fonts for non-Latin alphabet if we were to use Latin-only Inter on those pages. And this is another reason why I hesitate to include Inter in our dist files here. I'd suggest we merge this as-is and try it out with Cloud for now.

> Inter will appear much bigger at the same fontSize

Good thing this PR allows for a gradual adoption 😉

fk:
> We can try this in Gatsby Cloud, but probably not on .com static pages — didn't they use to include some translated content?

Good question. Pretty sure we don't currently do, the I18n stuff never was finished IIU/RC.

fk:
> I'd suggest we merge this as-is and try it out with Cloud for now.

💯 — let's get this in!

PR #410, opened 2020-10-14
Display activity statusText

julienp:
Show activity status text:

Screenshot 2020-10-19 at 13 50 34


src/components/BuildLogs/BuildActivityEntry.tsx
```diff
@@ -111,6 +120,12 @@ export function BuildActivityEntry({
           <span css={durationCss}>{formatDuration(duration)}</span>
         )}
       </div>
+      {showStatusText && (
+        <React.Fragment>
+          <div />
```
julienp, line 125: This is an empty grid cell which is used by the activity icon in the row above.

abhiaiyer91 (approved):

fk (approved):
LGTM, thank you Julien! 🙏

Elanhant (approved):

PR #412, opened 2020-10-14
fix(Modals): animation improvements for panels

Elanhant:
Changes related to panels:
* Replace keyframes animation for side panels with spring-based animation (`useSpring` from react-spring)
* Get rid of the horizontal scrollbar appearing when the side panel is in transition

Changes related to `StickyObserver`:
* Rename `setIsIntersecting ` to `setIsStuck` — I have no idea why I haven't done this from the start, the old naming is incredibly confusing when it comes to the use of `IntersectionObserver` where we used to check for the element _not_ intersecting and then calling `setIsIntersecting(true)` 🤦 
* Add an additional condition for "being stuck" — we are now checking an intersection entry for `intersectionRatio === 0 && !isIntersecting`. Without checking for `!isIntersecting` we'd get false positive "stuck" state in case the target is intersecting with the root _horizontally_ — we don't care about this case since `StickyObserver` is only supposed to be used for vertical sticky direction. To exclude this case, we are using the knowledge that `intersectionRatio` can be 0 when the area of the target's bounds rectangle is zero and `isIntersecting` is `false` ([source](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/intersectionRatio#Value))
* Make "lip" shadow more subtle instead of the current "shadow of doom" 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [0.0.217](https://github.com/gatsby-inc/gatsby-interface/compare/v0.0.216...v0.0.217) (2020-10-14)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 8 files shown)


fk (approved):
_[smooth](https://www.youtube.com/watch?v=sFvENQBc-F8)_ ✨ 🙏

PR #417, opened 2020-10-19
fix(icons): prevent plan icons from having weird black borders

Elanhant:
Adds missing `stroke="none"` prop to `IndividualPlanIcon` and `TeamPlanIcon` to avoid the inner dots getting surrounded by black borders which are inherited from current text color.


fk (approved):
🙏

greglobinski (approved):

PR #420, opened 2020-10-29
SidebarNav fix for inner page links

greglobinski:
# Purpose 

To fix issues with inner page linking (https://app.clubhouse.io/gatsbyjs/story/16993/the-scroll-to-section-isn-t-working-in-the-site-settings-side-bar) we have to use `a` tag instead of Gatsby `Link`  more context here https://gatsbyjs.slack.com/archives/CGJ457QTT/p1603890513003300 

Related PR https://github.com/gatsby-inc/mansion/pull/9916

### DIFF
--- a/src/components/SidebarNav/SidebarNav.tsx
+++ b/src/components/SidebarNav/SidebarNav.tsx
@@ -187,6 +187,15 @@ function SidebarBaseItem({
   href,
   ...rest
 }: SidebarBaseItemProps) {
+  const isOverwritten = to && href
+  const isText = (!to && !href) || isOverwritten
+
+  if (isOverwritten) {
+    console.warn(
+      `You can't set both 'to' and 'href' props on SidebarBaseItem at the same time! Check the '${label}' link.`
+    )
+  }
+
   return (
     <li
       css={theme => [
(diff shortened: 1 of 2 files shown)


fk (approved):

src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -202,9 +204,20 @@ function SidebarBaseItem({
...
+          href={href}
+          onClick={onClick}
+          css={itemLinkCss}
+          aria-current={current}
+        >
+          {label}
+        </a>
+      )}
+
+      {!to && !href && label}
```
sslotsky, line 220: If the parent provides both `to` and `href`, then what happens? Does the component just not display the label at all?

src/components/SidebarNav/SidebarNav.tsx
```diff
@@ -183,8 +184,18 @@ function SidebarBaseItem({
...
   onClick,
   current,
+  href,
   ...rest
 }: SidebarBaseItemProps) {
+  const isOverwritten = to && href
+  const isText = (!to && !href) || isOverwritten
+
+  if (isOverwritten) {
+    console.warn(
```
Elanhant, line 194: Let's use a `warn` helper function from `src/utils/maintenance/warn.ts` — it allows to only log these warnings in development mode.

Elanhant (approved):

PR #424, opened 2020-11-07
fix(form): fix styles for StyledRadio with multiline label

Elanhant:
This PR fixes the problem with our radio buttons that can be seen here:
Screen Shot 2020-11-06 at 5 52 00 PM

My first though was that the absolute positioning for the dot was off, and it indeed was — the offset was based on the parent's height (`100%`), which increases when the label takes more than one line. However, changing `100%` to the radio "circle" size was not enough:
Screen Shot 2020-11-06 at 5 51 34 PM

I found out that this occurs only when the screen is wide enough and when, coincidentally, we bump base font size on `html`. So I have set `font-size: initial` on the `StyledRadio`'s fake radio circle and somehow it worked 🤷‍♂️ 

Screen Shot 2020-11-06 at 5 51 13 PM


DSchau:
Love these write-ups @Elanhant!

fk (approved):
✨ Maybe when you're back from vacay Aleksei, the `html` base font size bump for larger screens will just be _gone_… 🐉 ⚔️ 😉 

🙏 <3

mfrachet (approved):

PR #425, opened 2020-11-07
chore: bring back color palette modal

Elanhant:
As the title says — this PR brings back the palette modal for each of our main colors. This modal can be opened by clicking on the color name in Colors story.

It will definitely need some polishing, but it should do its job for now 🙂 
localhost_6006_iframe html_id=theme-colors--page viewMode=story (4)


fk (approved):

fk:
<3

PR #429, opened 2020-12-04
feat(DropdownMenu): add support for 'tone' and 'Icon' props

Elanhant:
`DropdownMenuLink` and `DropdownMenuItem` now support two new props: `tone` and `Icon`. If this is approved and shipped, we can use those props to replace our `DropdownMenuActionLabel` component in Cloud.


fk (approved):
:shipit:

PR #430, opened 2020-12-05
feat(Text): support as='div' and margin reset props

Elanhant:
* Allows to use `div` in `Text`'s and `BaseText`'s `as` prop.
* `Text` now accepts `noMarginTop` and `noMarginBottom` props, which allow to reset browser defaults for the corresponding CSS properties.


fk (approved):
✅

greglobinski (approved):

PR #432, opened 2020-12-10
fix(form): set StyledTextArea's height to 'auto' so that 'rows' attribute works

Elanhant:
Currently we have to do some terrible things if we want to use `rows` attribute for text areas:
```jsx
<TextAreaFieldBlock
  rows={10}
  css={{ textarea: { height: `auto` } }}
/>
```

This was due to `StyledTextArea` inheriting styled input styles which have an explicitly set height. This PR fixes this oversight by setting `height: auto` for styled text areas.


fk (approved):
💯

PR #438, opened 2021-01-19
[fix] Preserve changing '_' to <em> in structured log error message

greglobinski:
CB: https://app.clubhouse.io/gatsbyjs/story/23024/structured-logs-formatter-interferes-with-error-message

Because we are using `Markdown` to format structured log messages. When the message contains something like 

```
Can't resolve '../lerna_node_version.json' in '/usr/src/app/www/lerna_version_node/src/pages'
```
the underscores inside the path strings are transformed to `<em>` tags as it's expected for markdown transformation so that results with such a view for a user.

Screenshot 2021-01-19 at 10 48 08

The PR prevents this by escaping `_` char if they are a part of a single quoted string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "scripts": {
-    "test": "jest --env=jest-environment-jsdom-sixteen",
+    "test": "jest --env=jest-environment-jsdom-sixteen --verbose",
     "test:a11y": "jest __tests__/AccessiblityStoryshots.test.js --testPathIgnorePatterns=[] --env=jest-environment-jsdom-sixteen",
-    "test:watch": "jest --watch --env=jest-environment-jsdom-sixteen",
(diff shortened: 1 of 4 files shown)


Elanhant (approved):
LGTM!

I wonder, however, if we could instead try to switch the quotes in such path-like strings to backticks ("\`")? This way they would be transformed into `<code>` tags and get the appropriate styles while also preserving the underscore (_) characters. It should be a relatively simple regex, I can help you to build it if you want. What do you think?

greglobinski:
> switch the quotes in such path-like strings to backticks ("`")

generally, I like the idea although I'm not sure if any single quoted string should be transformed into `<code>`, I will give it a try.

greglobinski:
I switched to `<code>` as you suggested, but I'm not truly satisfied :). I believe this need some general refactor, if it's possible, to introduce more consistency in rendering the messages. But for now it's enough. Thanks Aleksei.

PR #441, opened 2021-02-02
Unify paddings for all Notification variants

greglobinski:
There is no need for `SECONDARY` and `SOLID` `Notification` to have such wide paddings. 

Screenshot 2021-02-02 at 16 28 59

Let's unify the padding with the `PRIMARY` variant. 

Screenshot 2021-02-02 at 16 28 31

This will let reproduce design for the Hosting Features banners. 

Now the banner looks like below. 

Screenshot 2021-02-02 at 16 29 25

with the changes will look like here

Screenshot 2021-02-02 at 16 29 54

Take a look at the Design Doc point 2. 

https://www.figma.com/file/95HMgM1GfqL86C0Ff33hnr/Hosting?node-id=2622%3A733

### DIFF
--- a/src/components/Notification/Notification.helpers.ts
+++ b/src/components/Notification/Notification.helpers.ts
@@ -35,9 +35,6 @@ const variants: Record<NotificationVariant, GetVariantStylesFn> = {
       borderRadius: theme.radii[2],
       boxShadow: `0 0 0 1px ${theme.tones[tone].light} inset`,
       padding: `${theme.space[5]} ${theme.space[7]}`,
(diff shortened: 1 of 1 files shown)


gillkyle (approved):
Nice PR description, makes this very easy to approve 🙂

gillkyle:
I see some a11y check but that looks different

fk (approved):

greglobinski:
> I see some a11y check but that looks different
@gillkyle  Yep, for now a11y checks are broken, the check is not required to merge.

PR #446, opened 2021-02-16
ProgressBar extension - second value 

greglobinski:
## Purpose 

Add second optional value to render in `ProgressBar`, need to present `site` scoped hosting usage & limits at Cloud.

Screenshot 2021-02-17 at 00 10 42


fk (approved):
🚀

PR #468, opened 2021-06-23
chore(Tooltip): Adjust to spec, improve transition

fk:
TL;DR:

* non-breaking update: fixes a minor visual glitch, improves the look-and-feel and aligns the component with the spec, and links the documentation :v:
* preview: https://deploy-preview-468--gatsby-interface.netlify.app/?path=/story/tooltip--positions
* currently: https://gatsby-interface.netlify.app/?path=/story/tooltip--positions

### Details

- adjust Tooltip look and feel:
  - white instead of transparent white text `color`
  - reduce font-size from `14` to `12px`
  - increase `TooltipContent` padding from `4`/`8` to `8`/`12px`
  - don't overlap `TooltipPointer` and `TooltipContent` backgrounds — this increases the visual height of the `TooltipPointer` by `1px`, aligning it with the spec (`6px`), and improves the fade-in/-out transition animation during which the overlap is currently visible – ref. this screenshot of the tooltip during the transition/fade-in:  
    grafik
  - increase `MIN_WINDOW_SPACING` — the (currently horizontal only :/) offset to the viewport boundaries — from `4` to `8px`
- set `maxWidth` of `16rem`/`256px` for `TooltipContent`
- improve the Tooltip's `enter`/`leave` animation/transition:
  - up to now, the animation defined the Tooltip's start position "on top" of the trigger element ("10 pixels 'inside'"), and ending with the Tooltip positioned right at the edge of the trigger element
  - it also was not taking the Tooltip's position relative to the trigger element into account for the `translateY` part of the animation (defining the `position` part of the transition): the direction of the animated slide-in/-out of the Tooltip should depend on its placement relative to the trigger element (top/bottom)
  - **now:** note how the top Tooltip moves away from the trigger element when animating in, while the bottom Tooltip moves towards it:  
    recording
  - this PR changes the above so that the Tooltip's start position is right at the edge of the trigger element, and its end position is 4px away from the trigger element
  - the slide-in/-out animation direction is fixed
  - animation is slightly snappier
  - **this PR** (looks a bit weird because the GIF is cropped to loop at an unfortunate position):  
    this PR

Also links Figma spec/doc, and slightly adjusts stories.

[ch31933]


src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -20,6 +20,11 @@ export default {
     readme: {
       sidebar: README,
     },
+    design: {
+      type: "figma",
+      url:
+        "https://www.figma.com/file/WyBPGOvsiiElvusYrXQ6bY/Tooltip?node-id=19%3A4",
+    },
```
fk, line 27: Adds the Figma design/spec to the **Addons'** _Design_ tab: 
image

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -116,7 +121,7 @@ export const WithRichTooltip = Template.bind({})
 
 WithRichTooltip.args = {
   label: (
-    <span css={{ display: `inline-flex`, alignItems: `center` }}>
+    <span css={{ display: `inline-flex`, alignItems: `center`, gap: 4 }}>
```
fk, line 124: Add a little bit of whitespace in between `RichTooltip`'s icon and body copy, purely cosmetic: (before/after) 
image

image

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -131,6 +136,7 @@ export const WithLongTooltipText = Template.bind({})
 WithLongTooltipText.args = {
   label: LONG_TEXT,
   children: <button>Hover on me!</button>,
+  position: "bottom",
```
fk, line 139: Position `LongTooltipText` story tooltip at the `bottom` of the trigger element because we limit the tooltip max-width now, and the length of the demo copy would result in it not being fully visible using the default `top` position: (before/after)

image


image

src/components/Tooltip/Tooltip.tsx
```diff
@@ -47,11 +42,18 @@ export default function Tooltip({
...
-  const transitions = useTransition(
-    isVisible ? tooltipParams : null,
-    null,
-    transitionConfig
-  )
+  const transitions = useTransition(isVisible ? tooltipParams : null, null, {
+    from: { opacity: 0, transform: "translateY(0px)" },
+    enter: {
+      opacity: 1,
+      transform: `translateY(${position === "top" ? `-4px` : `4px`})`,
```
fk, line 49: The tooltip's enter and leave transitions now take the tooltip `position` into account – ditched `transitionConfig` while making that change.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -47,11 +42,18 @@ export default function Tooltip({
     ref: child.ref,
   })
 
-  const transitions = useTransition(
-    isVisible ? tooltipParams : null,
-    null,
-    transitionConfig
-  )
+  const transitions = useTransition(isVisible ? tooltipParams : null, null, {
+    from: { opacity: 0, transform: "translateY(0px)" },
```
fk, line 46: The transition now always starts (and ends, ref `leave`) at the outer edge of the trigger element, and not on top/"inside" of it.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -47,11 +42,18 @@ export default function Tooltip({
...
+    enter: {
+      opacity: 1,
+      transform: `translateY(${position === "top" ? `-4px` : `4px`})`,
+    },
+    leave: { opacity: 0, transform: "translateY(0px)" },
+    config: {
```
fk, line 52: Make the transition slightly snappier/faster.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -47,11 +42,18 @@ export default function Tooltip({
...
-    null,
-    transitionConfig
-  )
+  const transitions = useTransition(isVisible ? tooltipParams : null, null, {
+    from: { opacity: 0, transform: "translateY(0px)" },
+    enter: {
+      opacity: 1,
+      transform: `translateY(${position === "top" ? `-4px` : `4px`})`,
+    },
+    leave: { opacity: 0, transform: "translateY(0px)" },
```
wardpeet, line 51: Using translate3d should enable GPU acceleration which you should use when using transitions to lower redraws.

```suggestion
  const transitions = useTransition(isVisible ? tooltipParams : null, null, {
    from: { opacity: 0, transform: "translate3d(0, 0, 0)" },
    enter: {
      opacity: 1,
      transform: `translate3d(${position === "top" ? `0, -4px, 0` : `0, 4px, 0`})`,
    },
    leave: { opacity: 0, transform: "translate3d(0, 0, 0)" },
```

wardpeet (commented):
Small code change else it looks good from a code perspective :)

greglobinski (approved):
Good work 👍

fk:
<3<3<3

src/components/Tooltip/Tooltip.tsx
fk: Hm, applying this breaks the transform (opacity still working): https://deploy-preview-468--gatsby-interface.netlify.app/?path=/story/tooltip--with-accessible-label … taking a quick look.

fk: Resolved in a31454e :v:

PR #500, opened 2022-05-27
[feat] Icons - AgencyPlanIcon

greglobinski:
- add AgencyPlanIcon
- use "currentColor" instead of defined one


rmatambo8:
@greglobinski is this ready to merge?

PR #486, opened 2022-02-23
Inter for `<Heading>` and theme.fonts.heading

fk:
- "soft-deprecate" `<Heading fontVariant="UI" />`
- adjust design-token docs

PR #510, opened 2022-08-03
Inter for <Heading> and theme.fonts.heading

greglobinski:
It's a replacement of https://github.com/gatsby-inc/gatsby-interface/pull/486/files


aghreed:
I'm sure you're aware, but we've got a whole lot of `headingUI` usage in dotcom that could be cleaned up after this gets merged. Happy to help out with that.

